PR #4, opened 2021-04-22
[OPS-7438] Let us see if this works, just for the lols.

cafuego:


orakili (approved):
I've been working with the PHP 8 images for a while now and haven't encountered any issues.

PR #17, opened 2021-06-15
[RW-32] Rivers

orakili:
Ticket: RW-32

This ports the main rivers (lists of documents): /updates, /jobs, /training, /countries, /disasters and /organizations.


orakili:
@left23 Fyi, I've added a component in the subtheme for the icon sprite with a script to generate it from individual SVG icons from the Humanitarian Icon Set notably.

The script can extract colors from the CSS variables defined in `sass/cd/_cd-variables.scss` and in addition to the sprite generates a CSS file with a list fo CSS variables for use with the `background` property. Example:

```css
background: var(--rw-icons--common--help--12--dark-blue);
```

More details in the [README](https://github.com/UN-OCHA/rwint9-site/blob/afca0493cbce1c6a04ebd7167b68f35e0fe17c48/html/themes/custom/common_design_subtheme/components/rw-icons/README.md).

Let's discuss what you think of this approach when we have a chance.

PR #19, opened 2021-06-23
[RW-32] Country/disaster rivers

orakili:
Ticket: RW-32, RW-40, RW-42

This adds the country and disaster rivers and consolidate most of the templates, adding an "attributes" and "title_attributes" variables when relevant to help adding classes when overriding a template in the subtheme.

This also ensures the opportunities block on the homepage uses the same styling for its title (RW-40) and by adding the disasters river, the "view all disasters" is back on the homepage (RW-42)

### DIFF
--- a/html/modules/custom/reliefweb_entities/reliefweb_entities.module
+++ b/html/modules/custom/reliefweb_entities/reliefweb_entities.module
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Render\BubbleableMetadata;
+use Drupal\Core\Template\Attribute;
 
 /**
  * Implements hook_entity_bundle_info_alter().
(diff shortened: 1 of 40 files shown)


orakili:
@left23 Fyi, I added "attributes" and "title_attributes" to most templates so that we can more easily add the necessary classes in the template overrides in the subtheme. Example for the opportunities block on the homepage:

```twig
{% set title_attributes = title_attributes.addClass('cd-block-title') %}

{% embed '@reliefweb_homepage/reliefweb-homepage-opportunities.html.twig' %}

  {% block library %}
    {{ attach_library('common_design_subtheme/rw-opportunities') }}
  {% endblock %}

{% endembed %}
```

That makes me think we could possibly simplify the overrides as the above is equivalent to:

```twig
{{ attach_library('common_design_subtheme/rw-opportunities') }}

{% set title_attributes = title_attributes.addClass('cd-block-title') %}

{% include '@reliefweb_homepage/reliefweb-homepage-opportunities.html.twig' %}
```

It doesn't change much, so no strong opinion there. What do you think?

left23:
@orakili Yes, good suggestion.

PR #22, opened 2021-06-23
RW-31 Adjust template overrides to use twig include instead of embed

left23:
Remove twig blocks for library attachment in custom module templates
And replace the Twig embed with Twig include for sub theme template overrides


orakili (approved):
Looks good. Thanks.

PR #23, opened 2021-06-23
Rw 33 - Subscriptions

attiks:


orakili (approved):
@attiks Thanks. I notably tested the drush functions and compared the emails with the ones generated by the D7 site.

I have added some changes and created follow-up tickets:

**Changes**

- Simplified the templates so that we don't need to override all of them in the common design subtheme. Only the header template is overriden to add the logo.
- Added a twig filter for the CSS inlining and using it in the `mimemail--message` template.
- Fixed some issues with the template titles.
- Use the ThemeHandler to get the default theme rather than having `common_design_subtheme` hardcoded in the module.
- Moved the css to the `rw-subscriptions.css` component so that the CSS can be re-used for the email previews.
- Removed the PHPMailer SMTP module to use the built-in mail() so that `sendmail` (msmtp) is used. This is the same way the D7 site operates with the mail host/port/credentials managed by ansible. [Discussion with Cafuego](https://www.flowdock.com/app/reliefweb/infrastructure/threads/Y6oSrGJG9OOnt8wTm3ZVrI99yNb).

@attiks Regarding the SMTP changes. According to the [doc](https://docs.docksal.io/service/other/mailhog/#php-configuration), you probably need to add the following in the `.docksal/etc/php/php.ini` file:

```
; Mail settings
sendmail_path = '/usr/local/bin/mhsendmail --smtp-addr=mail:1025'
```

**Follow-up tickets**

- [RW-79] Add the admin overview page (`/admin/subscriptions/overview`) which provides a list of all the subscriptions with the number of subscribers, timestamps and previews
- [RW-81] Add route and page to unsubscribe (logged in users are redirected to the subscriptions page, anonymous users are provided a form to unselect their subscriptions)
- [RW-85] Move most of logic from `ReliefwebSubscriptionsSendCommand` to a service so that queueing methods for example can be called when updating an entity.
- [RW-86] Review the logic for queueing the triggered notifications. `queueTriggered` for example is currently not called.
- [RW-87] Port `subject callbacks` used to generate the email subject of disaster and OCHA sitrep notifications
- [RW-88] Review the notification definitions, notably the API payload. For example the disaster API payload doesn't match the D7 version (missing status filter).


There are also some extra tickets (RW-84, RW-89, RW-90) for improvements but they are for after the launch.

Merging to move forward.

PR #26, opened 2021-06-29
[RW-34] Add class for core meta

orakili:
Ticket: RW-34

This adds a flag to the "entity-meta" template to add a class to distinguish the "core" meta information in river articles and below the title on article pages from the "full" meta information in the sidebar of the article pages.


left23 (approved):
Thanks

PR #28, opened 2021-06-29
RW-56 Disasters

left23:
Add attributes variable to the description element to add necessary attributes for dropdown javascript.

Page title component with Disaster icon and  rw-meta.

Coding standards - Remove an extra space from `#main-content .rw-river-article--disaster[data-disaster-type="WF"] .rw-river-article__title a:before` selectors

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-entity-meta/rw-entity-meta.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-entity-meta/rw-entity-meta.css
@@ -161,7 +161,7 @@
   background: #f65c51;
 }
 
-.taxonomy-term--disaster--full dl.rw-entity-meta dd.rw-entity-meta__tag-value--status--past:before {
+#main-content .taxonomy-term--disaster--full dl.rw-entity-meta dd.rw-entity-meta__tag-value--status--past:before {
   content: none;
 }
 
--- a/html/themes/custom/common_design_subtheme/components/rw-entity-meta/rw-entity-meta.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-entity-meta/rw-entity-meta.css
@@ -161,6 +161,10 @@
   background: #f65c51;
 }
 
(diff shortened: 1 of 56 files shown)


html/themes/custom/common_design_subtheme/components/rw-entity-meta/rw-entity-meta.css
```diff
@@ -161,26 +161,17 @@
   background: #f65c51;
 }
 
+.taxonomy-term--disaster--full dl.rw-entity-meta dd.rw-entity-meta__tag-value--status--past:before {
```
left23, line 164: @orakili Can you determine why the Disaster status isn't being applied here?
The `<dl>` gets a class of `rw-entity-meta__tag-value--status--ongoing` even though it should be `past`
Then this rule would work.
Example:

orakili: @left23 The disaster status entity meta class in `taxonomy-term--disaster--full.html.twig` was hardcoded to `ongoing`. Fixed in https://github.com/UN-OCHA/rwint9-site/pull/28/commits/a214750e8b77ad8d633c3a54f3d10a7e1abbb948.

orakili (dismissed):
Looks good to me. There was missing content for the Affected countries. It's tackled in https://github.com/UN-OCHA/rwint9-site/pull/36.

left23:
@orakili Thanks.
I'm seeing a discrepancy between the Disaster status in the River `/disasters` and the Disaster page itself.
See Malaysia: Floods - May 2021 as an example `disaster/fl-2021-000061-mys`. On the river I see it as a past disaster, but on the page, it's ongoing.
Do you get the same result?

orakili:
@left23 The data in the river comes from the API so it has the latest version while the individual pages (country, disaster etc.) use the database data. This disaster was changed to past disaster on June 29 so what you see is normal. On prod, there will not be such inconsistencies because the DB and API will be in sync.

left23:
@orakili I'm not sure why it's asking for a review again. I've only updated this PR with one commit.

orakili (approved):

PR #32, opened 2021-06-30
[RW-59] Style active letter navigation link

orakili:
Ticket: RW-59

This adds back a piece of [javascript](https://github.com/UN-OCHA/rwint-site/blob/bfb100de099df2c02a2319e7e2b5af0e3c12b484/site/profiles/reliefweb/themes/kobe/js/site.js#L12-L28) that was used to style the active link in the /countries river page. 

It also updates the CSS rule to style this active link which was before using [data-active] but, now, uses `rw-river-letter-navigation__link--active`.

Finally a tiny change to center vertically the letters.

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-river-letter-navigation/rw-river-letter-navigation.js
+++ b/html/themes/custom/common_design_subtheme/components/rw-river-letter-navigation/rw-river-letter-navigation.js
@@ -6,7 +6,7 @@
    */
   function updateActiveFragmentLink() {
     var hash = location.hash || '#';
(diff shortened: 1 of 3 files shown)


html/themes/custom/common_design_subtheme/components/rw-river-letter-navigation/rw-river-letter-navigation.js
```diff
@@ -0,0 +1,27 @@
+(function () {
```
orakili, line 1: Note: I'm not adding `Drupal` because technically this scripts doesn't require it.

left23 (approved):

PR #38, opened 2021-07-01
RW-56 Layout rules for advanced search and rivers list

left23:
- Add layout rules for advanced search and rivers list, update selectors.

There's still no vertical rhythm and I'm not sure if the generic `rw-river` css file is appropriate for the composition rules. It's a fine place for it in the meantime.

Currently, the Countries and Organizations river pages (maybe Topics also?) don't include the advance search, so the content column spans full width.
But the Organizations page shares the generic `reliefweb-rivers-page.html.twig` template, whereas the Countries page `reliefweb-rivers-country-list.html.twig` has its own template. 

The class `rw-river--river-list` is being used as a selector for the layout CSS. The Country page doesn't have this class so I've used `rw-river-page:not(.rw-river-page--organizations) .rw-river--river-list` to omit Organizations only.

### DIFF
--- a/html/modules/custom/reliefweb_rivers/src/RiverServiceBase.php
+++ b/html/modules/custom/reliefweb_rivers/src/RiverServiceBase.php
@@ -292,6 +292,11 @@ abstract class RiverServiceBase implements RiverServiceInterface {
   public function getRiverAdvancedSearch() {
     $advanced_search = $this->getAdvancedSearch();
 
+    $settings = $advanced_search->getSettings();
+    if (empty($settings['filters'])) {
+      return [];
+    }
+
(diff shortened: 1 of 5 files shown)


orakili (approved):
@left23 Thanks. I've added classes for each feature present in the river page to help with the styling:

- `rw-river-page--with-views`
- `rw-river-page--with-search`
- `rw-river-page--with-letter-navigation`
- `rw-river-page--with-advanced-search`

I replaced the `:not` rule using the `rw-river-page--with-advanced-search` class as the blog and topics rivers also don't have filters.

Hopefully those classes will also help with the vertical rhythm.

PR #39, opened 2021-07-06
RW-39 Country slug in Headlines

left23:
- update selector so country slug is correctly sized in headlines component


orakili (approved):
Thanks.

PR #46, opened 2021-07-23
Rw 73

attiks:


orakili (approved):
Thanks. It's working well.

I added the following:

- template for the link so that it could be easily styled using a component as usual
- `rw-bookmark-link` component to style the above link as on the D7 site
- `bookmark content` permission
- entity status conditions to limit bookmarking to published entities
- entity update and delete hooks to remove bookmarks when an entity is unpublished or deleted
- moved the pager to the reliefweb_rivers_river theme which already was handling that
- added results at the top of the paginated lists
- added message for when the list are empty (applies to other places using the river theme)
- added bookmark link to training pages

Merging. No follow-up tickets.

PR #47, opened 2021-07-23
RW-78 Full node styles

left23:
Jobs, Reports, Training, Book pages, Blog posts in `rw-document` component
`rw-book-menu` and `rw-related-content` components


left23:
@orakili This is ready for review. 
Vertical rhythm rules are still missing.
Also note the borders on the Related Content are not quite right. The rules to control the border are present in both `.rw-blog` and `.rw-headlines` components. I will deal with this step when I start removing parent selectors from the rules and see where the conflicts or re-usability occur.

orakili (approved):
Thanks. I've made some adjustments and fixed a few non-styling issues that impacted those node pages:

- The rules to style the meta information in the sidebar were not applied
- The training feed information section was missing
- The related content library was not attached to the training pages
- Removed the special styling of the book menu heading to be consistent with the other headings in the sidebar (ex: latest blog posts)
- Fixed the class for the content format for reports

There is a follow-up ticket to discuss the styling of the "fee information" and "how to apply/register" sections in the job/training pages: RW-92

orakili (approved):

PR #51, opened 2021-07-28
[RW-73] Bookmarks overview page

orakili:
Ticket: RW-73

@attiks This PR contains a working example for the content of the "My bookmarks" page using the data from the API.

**Note**: the actual code for this PR is in this commit: https://github.com/UN-OCHA/rwint9-site/commit/0ed116fa86c1d069ae432aa4f4acbe21ecf6112a, the rest is just from merging develop into the `RW-73` branch.

Basically, for each type of bookmark (report, job etc.), it retrieves the entity ids from the bookmarks table and add those ids as a filter to the API query for the corresponding API resource.

Currently, it shows the 5 more recent items for each bookmark type on the same page as an example:

Screen Shot 2021-07-28 at 13 55 16

However what we'd want is actually "sub-pages" (tabs) for each bookmark type with a paginated list of items and maybe have the "My bookmarks" landing page just show the number of bookmarks for each category linking to the sub pages.

The pagination can be done on the SQL query to get the IDs from the reliefweb_bookmarks.

Let me know what you think.


attiks (approved):
Thanks!

PR #55, opened 2021-08-05
[RW-106] Add river article classes

orakili:
Ticket: RW-106

This is to illustrate a way to add classes to river articles so that different styles can be applied (like headline, maps, appeals).

**Note**: there are already some classes to distinguish between articles with a summary, thumbnail etc., ex: `rw-river-article--with-summary`, which in the card: https://trello.com/c/uYummb5u was referenced as `compact`.

We could extend this article class approach and add a `meta_attributes` variable to the `reliefweb_rivers_river_article` theme and in the `reliefweb-rivers-river-article--report.html.twig` template override, we could add a class for the meta information based on the article class.

For example to add class to always show the entity meta info for the maps section on country page as stacked instead of inline:

```
{% if attributes.hasClass('rw-river-article--map') %}
  {% set meta_attributes = meta_attributes.addClass('rw-entity-meta--stacked') %}
{% endif %}
```

@left23 Let me know what you think and if you want I can extend this PR with the meta attributes.

### DIFF
--- a/html/modules/custom/reliefweb_entities/reliefweb_entities.module
+++ b/html/modules/custom/reliefweb_entities/reliefweb_entities.module
@@ -149,6 +149,8 @@ function reliefweb_entities_theme() {
         'core' => FALSE,
         // Meta information, see "reliefweb_entities_entity_meta" above.
         'meta' => '',
+        // Meta information attributes.
+        'meta_attributes' => NULL,
       ],
     ],
     'reliefweb_entities_entity_social_media_links' => [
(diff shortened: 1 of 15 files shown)


left23 (dismissed):
@orakili Looks good, let's try it.

orakili:
@left23 I extended the PR with attributes for the meta information and an example in the `reliefweb-rivers-river-article--report.html.twig` override to add "rw-entity-meta--stacked" class for the river articles appearing in the `Maps and Infographics" and "Appeals and Response Plans" section.

Also added the meta attributes to the `reliefweb_entities_entity_details` theme so that we can also easily add a class to style the meta information in the `node` templates.

I think with both the `article_attributes` and `meta_attributes`, we should be able to get remove all the parent selectors.

PR #57, opened 2021-08-06
[RW-80] Global search

orakili:
Ticket: RW-80

This adds the search to the main header and the search results page (/search/results) which shows a list of resources (ex: reports, countries) matching the search query.

Note: for countries, the full list of matching countries is displayed as the country river is not searchable.


orakili:
@left23 This doesn't include the styling of the search results page. Ticket for that is RW-108.

If you don't find any problem and are happy enough with the implementation, please merge.

left23 (approved):

PR #58, opened 2021-08-09
RW-31 Refactoring continued

left23:
1. RW-51 Images for Appeals and Response plans, separate rules from generic article image rules
2. RW-61 Results styles for Blog list and make rule more generic
3. RW-68 Remove page title styles for Views lists
4. RW-104 Correct entity meta styles for organisations
5. Comment out rule for body font-size to use base theme instead, not sure what knock-on effects this might have for the other typographic elements.
6. RW-92 Training/job pages: review styling of fee information and how to apply/register sections
7. RW-58 River: search - space around search bar
8. Additional style tweaks based on visual QA
9. RW-53 Country page: Key figures links on separate lines
10. RW-52 Country page: sidebar TOC overlaps with CD footer (might still need tweaking as this is related to the sticky position and the 100% height added to all the parent divs)


left23:
@orakili Please keep in mind that this PR https://github.com/UN-OCHA/rwint9-site/pull/60 is a follow up to add vertical rhythm so there may be remaining issues in this PR that is resolved in the other.

orakili (approved):
Looks good, thanks.

Some issues:

**Full article meta in header on small screens**

country-page-meta-separator

On full node/term pages (ex:  country page, report page), on small screens the separator after the last `dd` appears in the meta information below the page title. On larger screen, lines 76-78 or rw-entity-meta.css hide it. 

I don't think of a use case where the separator would need to be displayed for the last `dd` so I would suggest something like:

```
.rw-entity-meta--core .rw-entity-meta__tag-value--last:after {
  content: none;
}
```

**Organization details - font size**

org-page-details

Due to the removal of the `body { font-size: 16px: }` (which makes sense), some elements like the links in the `Details` section of organization pages inherit the default font size but should be smaller (16px for example).

**View more links - font size**

view-more-large-font

Same as above for the org. details.

left23:
@orakili I've refactored the meta rule again, using `--core` class to set the inline display on desktop, and fixed the other items.

orakili (approved):
Looking good, thanks! Merging.

PR #60, opened 2021-08-11
RW-31 vertical rhythm

left23:
RW-55

This PR attempts to address the vertical rhythm in general by using the `cd-flow` component in place of more component-specific margin and padding rules. Because the `cd-flow` add a margin to the top of elements, I've reversed a lot of the rules where previously there was a top border and top padding and adjustments for the first-child, now there is a bottom border, and adjustments to the last-child.

There are also changes to rules in media queries as I start to check everything across multiple viewport dimensions.


left23:
@orakili I'm adding you as reviewer but let's not merge this until we can chat.

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
```diff
@@ -21,15 +25,19 @@
   position: relative;
   display: flex;
   flex-direction: column;
-  padding: 24px 0;
-  border-top: 1px solid #e6ecef;
+  padding: 0 0 24px 0;
```
orakili, line 28:
```suggestion
  padding: 0 0 var(--cd-flow-space) 0;
```

With this change the padding would have the same value as the margin on the other side of the border. What do you think?

html/themes/custom/common_design_subtheme/components/rw-blog/rw-blog.css
```diff
@@ -20,7 +20,12 @@
 }
 
 .rw-river-article--blog-post {
-  padding: 12px 0;
+  --cd-flow-space: 12px;
+
+  padding: 0 0 12px;
```
orakili, line 25:
```suggestion
  padding: 0 0 var(--cd-flow-space);
```

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
```diff
@@ -4,20 +4,27 @@
  * - tighter list
  * - hidden meta information
  */
+.rw-river--disasters {
+  --cd-flow-space: 1rem;
```
orakili, line 8: The `1rem` differs from the other places overriding the `--cd-flow-space` which use `px` and actually it doesn't seem to be used at the `.rw-river--disasters .rw-river-article--disaster` below override the margin-top (line 13).

Should this be changed to `16px` and line 13 removed and the padding changed to `padding: 0 0 var(--cd-flow-space)` below?

Also same comment as above about the issue with the `css-vars-ponyfill` not picking up the css var override.

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
```diff
@@ -4,20 +4,27 @@
...
+  /* border-top: 1px solid #e6ecef; */
+}
 .rw-river--disasters .rw-river-article--disaster {
   position: relative;
   margin: 16px 0 0 0;
```
orakili, line 13: This overrides the margin from the `cd-flow`.

orakili (commented):
Just a first pass at reviewing the code with some comments related to the use of the `cd-flow`. Some other generic thoughts (not particularly for the RW work):

I find that the use of the `cd-flow` makes the components less self-contained as they somehow need to be aware of where the `cd-flow` class was added and have to override the `--cd-flow-space` variable and the styling can be off if the `cd-flow` class is not added.

Also the `cd-flow` class only deals with the margin top but in all the cases it applies here, there is also a matching padding bottom. A solution is to use  `padding-bottom: var(--cd-flow-space)` and I think that should be added to the `cd-flow` component as a variant like `--cd-flow--top-bottom` (with a better name probably).

Another thought: I'm wondering if rather than adding the `cd-flow` classes to the templates, it would not be better to simply use the `--cd-flow-space` variable directly for the margins and paddings in the components to ensure consistent vertical rhythm. This way the components would be more self-contained but with consistent spacing, similar to using the color and font variables.

Note: I think we should add a section in the [CD README](https://github.com/UN-OCHA/common_design/blob/main/README.md) to explain how to use the `cd-flow` and `cd-bumper` classes if it's meant to be the standard way to control vertical rhythm.

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
```diff
@@ -21,15 +25,19 @@
...
+  border-bottom: 1px solid #e6ecef;
+}
+.rw-river-article:last-child {
```
orakili, line 31: Removing the bottom padding affects sections like the maps and infographics on the country/disaster pages:

maps-cards

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --cd-flow-space: 24px;
```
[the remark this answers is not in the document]
left23: I think this works. The main place where setting the flow value is useful is for the rivers - so the section title and the following elements `.rw-river__articles` and its children have the same vertical spacing.
In the context where we want a different value, I can define the flow value on a parent element, which as you said will be ignored by the polyfill so in IE11, the value on root will be used.

I am trying the `cd-flow` component here because all the values I mention are already 24px. I don't think variants are needed in this case.

html/themes/custom/common_design_subtheme/components/rw-blog/rw-blog.css
```diff
@@ -20,7 +20,12 @@
 }
 
 .rw-river-article--blog-post {
-  padding: 12px 0;
+  --cd-flow-space: 12px;
```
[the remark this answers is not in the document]
left23: True.

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
left23: I don't think there is a need to mix the function of the `cd-flow` here. Padding is about what happens within a component, the flow is about how the component interacts with other elements around it. I don't mind using variables for padding but I am against using the same variable because they are not the same thing.

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
left23: In this case 1rem = 16px which is why I didn't differentiate. I can make the units consistent nearer the end of the project when I replace values with variables.
I've removed the margin declaration so this uses the cd-fow value.

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
left23: I'm not sure if you mean `cd-flow` from the base theme? It appears to be working the way I intended. Am I misunderstanding what you're pointing out?

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
left23: fixed

left23:
@orakili I've responded to the comments, made some changes, and more commits to do with the vertical rhythm and general QA.
This is ready for another review.

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
orakili: Fair enough.

html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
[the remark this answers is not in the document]
orakili: Ok. I would be in a favor of adding a note about that in the common_design_subtheme's README for future maintainers. Making a note for myself for now and I can do that when I do a documentation update of the modules/themes later.

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
orakili: Roger that.

html/themes/custom/common_design_subtheme/components/rw-homepage/rw-homepage.css
orakili: II mentioned that already in https://github.com/UN-OCHA/rwint9-site/pull/60#discussion_r689958035 and you changed it. It was just to point out that the `margin-top` from the `cd-flow` class didn't apply because of this `margin` rule.

orakili (commented):
Looking good. I just noticed some inconsistency regarding the spacing around the river results (showing X of Y). Hopefully the individual comments below are clear enough.

html/modules/custom/reliefweb_rivers/templates/reliefweb-rivers-river.html.twig
```diff
@@ -29,17 +31,21 @@
...
+    resource ? 'rw-river--type-' ~ resource|clean_class
   ])
 }}>
   <h{{ level }}{{ title_attributes.addClass([
     'rw-river__title',
   ]) }}>{{ title }}</h{{ level }}>
 
   {% if entities is not empty %}
-    {{ results }}
+    {{ results|merge({
```
orakili, line 42:
```suggestion
    {{ results ? results|merge({
```

@left23 Sorry, there was a bug in the https://github.com/UN-OCHA/rwint9-site/pull/67 PR, this suggestion fixes it.

html/themes/custom/common_design_subtheme/components/rw-related-content/rw-related-content.css
```diff
@@ -33,10 +31,15 @@
...
+    margin-left: 16px;
+  }
+  .rw-document + .rw-river--type-reports .rw-river__articles article:nth-last-child(2) {
+    padding-bottom: 0;
+    border-bottom: none;
   }
-  #main-content .rw-document + .rw-river--type-reports .rw-river__articles article:nth-child(-n+2) {
-    border-top: none;
+  .rw-document + .rw-river--type-reports .rw-river__articles article:nth-child(2) {
+    margin-top: 0; /* Compensate for cd-flow top margin */
```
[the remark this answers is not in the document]
left23: Ok, good point. I will try to consolidate those rules as they only thing unique is the selectors.

I'm not sure I understand your comment about the variants, or rather, I am not sure how I should achieve this.
For the Blog and Organizations, their River sections has the ID of `river-list` which I think is too generic (as it is also used on the pages with the Advanced search) and doesn't differentiate between the list versus card type layout. Am I missing something?

The Related Content block seems to be missing an ID which might explain why I've used `.rw-document + .rw-river--type-reports .rw-river__articles` instead of something more direct.

In the meantime, I've updated the Organizations page to have the same treatment for the cards as the Blog page.

html/themes/custom/common_design_subtheme/components/rw-river/rw-river-results.css
```diff
@@ -3,6 +3,7 @@
  */
 .rw-river-results {
   width: 100%;
+  padding: 13px 0 0;
```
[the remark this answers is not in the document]
left23: Done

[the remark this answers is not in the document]
orakili: Looking good, thanks.

html/themes/custom/common_design_subtheme/components/rw-related-content/rw-related-content.css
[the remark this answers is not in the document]
orakili: Hmm, for the blog and organizations, I think we can add something like the following snippet in the `reliefweb-rivers-page.html.twig` template override in the subtheme to add a class to the river list wrapper:

```
{% if river == 'organizations' or river == 'blog' %}
  {% set content = content|merge({
    '#attributes': create_attribute().addClass('rw-river--river-columns'),
  }) %}
{% endif %}
```

Then if needed in the `reliefweb-rivers-river.html.twig` template override we could add a class to the articles if the `rw-river--river-columns` (or whatever class name is chosen) is present in the `attributes`.

For the Appeals and Maps, I guess we could check on the ID.

Same for the Related content actually after adding the`#id` after the line to https://github.com/UN-OCHA/rwint9-site/blob/1ca75688bf052c135d888d0b3b9105a52a2b2f8f/html/modules/custom/reliefweb_entities/src/DocumentTrait.php#L201

What do yo think?

orakili (commented):
I checked all the different types of front-end pages and they look good. 

The cd-flow approach seems to be working well. 

Is there anything in particular you'd like me to review in more details?

What else do you think needs to be done regarding the vertical rhythm?

left23:
@orakili If this is good to go, let's merge it. 

The consolidation of Blog, Organizations, Related, Appeals, and Maps isn't forthcoming because the layouts differ across viewports
Eg. Appeals displays as one column at 768px, and the others display as two columns. On desktop, Maps has 3 columns and the others have two columns.
I'll put my attempts in another PR.

orakili (approved):
Merging.

PR #62, opened 2021-08-17
[OPS-7696] New PHP images

cafuego:
Bump to the latest (PHP 8.0.9) version in Alpine 3.14, with an nginx DNS resolver fix and the crufty wddx.ini config file removed.


orakili (approved):
Working well. Thanks.

PR #65, opened 2021-08-18
[RW-81] Error/exception fixes

orakili:
Ticket: RW-81

PR against the RW-81 branch which fixes 2 issue found when testing sending notifications:

- https://github.com/UN-OCHA/rwint9-site/commit/51e7b3a28bc1806135c1faabd96a22f24aa56804: Fix issue with the subscriptions:send command when there are no queued notifications
- https://github.com/UN-OCHA/rwint9-site/commit/bfbebc44cb72bef0087e7b87de2356e5d3a0b934: Prevent exception when sending notifications due to duplicated Reply-To header

Note: I also merged develop into RW-81 before thus all the commits before the 2 above.


attiks (approved):
Checked both commits, looking good

PR #66, opened 2021-08-18
RW-109: XML sitemap

attiks:


orakili (approved):
Thanks, working well.

I updated the logging so that messages appear in syslog etc. There was also an issue with the url aliases because in D9, paths and aliases start with a `/`.

I also added nginx rules to handle the `/sitemap.xml` and `/sitemap.xsl` routes.

PR #67, opened 2021-08-19
[RW-31] Add attributes for the river results

orakili:
Ticket: RW-31

This adds attributes for the results in the river template and a class variant when in a river with filters in the sidebar to help styling the results: `rw-river-results--with-advanced-serarch`.


left23 (approved):
Looks good, thanks

PR #69, opened 2021-08-23
RW-31 column layouts

left23:
This is an attempt to consolidate the layout for

1. Blog list
2. Organizations list
3. Appeals and Response Plans block on Country page
4. Maps and Infographics block on Country page
5. Related block on Report page

As per this comment https://github.com/UN-OCHA/rwint9-site/pull/60#discussion_r691778092

The difficulty is the lack of shared layout rules these elements have, as explained here: https://github.com/UN-OCHA/rwint9-site/pull/60#issuecomment-903811208

### DIFF
--- a/html/modules/custom/reliefweb_entities/src/DocumentTrait.php
+++ b/html/modules/custom/reliefweb_entities/src/DocumentTrait.php
@@ -199,6 +199,7 @@ trait DocumentTrait {
 
     return [
       '#theme' => 'reliefweb_rivers_river',
+      '#id' => 'related',
       '#title' => $title,
       '#resource' => 'reports',
       '#entities' => $entities,
--- a/html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-article/rw-article.css
@@ -119,23 +119,33 @@
     /* Compensate the margin from the articles. */
     margin: 0 -8px;
   }
-  .rw-river--river-columns .rw-river__articles .rw-river-article {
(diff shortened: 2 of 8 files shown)


left23:
@orakili This is ready for review.

I'm not sure it offers much (anything?) since we still end up with rules like https://github.com/UN-OCHA/rwint9-site/pull/69/files#diff-d03a1ca59e91356fd162b5c0369588528efeb22e349be2535699637e777c16feR148 and https://github.com/UN-OCHA/rwint9-site/pull/69/files#diff-1b4af2a2d4b14bf6bf9462bfec1beb2ba1d2072d0ee657fe931e511c6fafc6c7R105-R131
so despite trying to use classes like `rw-river--river-columns--two` we still end up with rules specific to a block type (i.e. I don't think we can assume that `rw-river--river-columns--two` or `rw-river--river-columns--three` would work in other scenarios because while these blocks all share styles, there are enough differences to have exceptions.

In the same way, I attempted to add `rw-river--river-columns--three` for the Maps block but removed it as it didn't make sense because the padding is so specific to the individual article markup for Map items and would likely not apply to another scenario where a 3 column layout is used (another scenario doesn't exist).

Anyway, see what you think. I have no preference.

orakili:
@left23 Thanks. I tested and it seems to work well but as you mentioned I'm not sure that brings much. If that's ok, I'd like to keep this PR open a bit longer as I'd like to test a couple of things when time allows.

orakili (approved):
All good and merging as it makes styling of the topics landing page easier (#97).

PR #59, opened 2021-08-09
RW-31 admin theme

left23:
I've merged https://github.com/UN-OCHA/rwint9-site/pull/74 into this branch for the Theme Switcher module.

Also includes:

- local tasks styles
- Form styles (currently added as part of global library)
- Template overrides for Node edit forms and Term edit forms started - general method arrived at from the discussion on #74 , most markup present but specific fields and general details have not been covered (i.e. everything is a mess)
- Admin menu component (menu items currently hard coded)
- RW Headlines component with JS in place (needs to be ported to function)
- User pages started - templates in place and some styles copied from D7 (Subscription form, Dashboard template)

Currently using https://github.com/UN-OCHA/common_design/pull/254 base theme branch `cd-304-forms`

PR #74, opened 2021-08-31
[RW-117] Theme Switcher module

orakili:
Repurposed to install, enable and configure Theme Switcher module.

Below is the background on the initial exploration of how to override the Taxonomy term edit form templates.

<hr>

Ticket: RW-117

An example of how custom forms with grouped fields could be handled.

This overrides the `form.html.twig` template with an extra condition: check if a template matching the form id (attribute so that it's already in the form `xxx-yyy-zzz`), in which case render it otherwise render the `children` as usual.

In the template matching the form id, fields are rendered separately using the `element` variable (see [template_preprocess_element](https://api.drupal.org/api/drupal/core%21includes%21form.inc/function/template_preprocess_form/9.0.x)) and there is a loop at the end to ensure all the form elements are rendered.

**Note**: for these templates to be used, the form needs to use the CD theme/subtheme. The [theme_switcher](https://www.drupal.org/project/theme_switcher) contrib module allows to select which theme and admin theme to use per vocabulary, content type, path etc.


left23:
@orakili @attiks I've checked this out locally, and added to this PR to install, enable and config `theme_switcher`.
I can see it working for Nodes  eg: `/node/3758032/edit` but it doesn't seem to work with Terms eg. `/taxonomy/term/13/edit`

The Appearance setting is set to use `seven` and this is what I see for the taxonomy form. Can you have a look and see if you can identify where I am going wrong?

orakili:
@left23 I think you are hitting this problem: https://www.drupal.org/project/theme_switcher/issues/3123722

I think you can get around that by creating different rules instead of adding several conditions to only one. So one set of rules for nodes, one for terms, one for special paths.

left23:
@orakili I've merged this branch into https://github.com/UN-OCHA/rwint9-site/pull/59 so I can test more holistically, and it all seems to work well.

We have the generic term template override in D7 https://github.com/UN-OCHA/rwint-site/blob/master/site/profiles/reliefweb/themes/kobe/templates/forms/term-edit.tpl.php

There is no template suggestion in this case so for now I created an unused `taxonomy-term-form.html.twig` and I can extend that with the more specifically-named templates like in this commit https://github.com/UN-OCHA/rwint9-site/commit/220cbbbf98230775c14e393cf45f219b89237fd8
I'm not sure if this is the way to go (I would need to do this for each vocabulary) so I'll wait until we discuss it.

You mentioned earlier about potential limitations and to use https://github.com/UN-OCHA/unocha-term-front/blob/main/unocha_term_front.module as a reference for theme suggestions. When I apply the theme suggestion hook, I get the appropriate suggestions but they don't take effect. Can you reiterate what you said earlier about this if it's relevant?
I think I encountered a similar issue on my first attempt. See @attiks comment here https://humanitarian.atlassian.net/browse/RW-117?focusedCommentId=129203
I'm not clear whether `theme_switcher` module is intended to correct this for us.

Eg:
```
 FILE NAME SUGGESTIONS:
   * form--taxonomy-term-career-category-form.html.twig
   * form--taxonomy-term-form.html.twig
   x form.html.twig
```
Even though I have `form--taxonomy-term-form.html.twig` in place, it is not picked up.

orakili:
@left23 The suggestions are not picked up because they contain dashes instead of underscores (I've been bitten by that a few times... see explanation at the bottom of this comment if interested).

So, the following works:

```
/**
 * Implements hook_theme_suggestions_HOOK_alter().
 */
function common_design_subtheme_theme_suggestions_form_alter(array &$suggestions, array $variables) {
  $route_match = \Drupal::routeMatch();
  if ($route_match->getRouteName() === 'entity.taxonomy_term.edit_form') {
    
    $suggestions[] = 'form__taxonomy_term_form';
    // Suggestion in the form "form__taxonomy_term_VOCABULARY_form".
   // @todo remove once CD base theme's "suggestions_form_alter" hook is fixed.
    $suggestions[] = 'form__' . $variables['element']['#form_id'];
    // Move the more generic suggestion at the top of the suggestions to ensure
    // the suggestion with the vocabulary takes precedence if there is a
    // corresponding template.
    array_unshift($suggestions, 'form__taxonomy_term_form');
  }
}
```

Note that the CD theme needs a fix: https://github.com/UN-OCHA/common_design/blob/3ffe3bf6b08bc54e89bfb0039c1b69761002fbe4/common_design.theme#L273 `['#id']` should be changed to `['#form_id']`.

**What happens**

Suggestions with dashes appear in the twig debug info because it's generated from the `$suggestions` array  (not from the templates).

However, during template discovery Drupal looks for all the potential templates in the enabled modules and themes and store the discovered templates in an array with the filenames without extension **and** with dashes converted to underscores as keys.

For example if you have a template named `common_design_subtheme/templates/form/form--taxonomy-term-form.html.twig` then this will be stored as `$templates['form__taxonomy_term_form'] = $template_path`.

So when Drupal analyzes the theme suggestions, it does something akin to

```
foreach ($suggestions as $suggestion) {
  if (isset($templates[$suggestion])) {
    return $templates[$suggestion];
  }
}
```

In that case if the suggestion uses dashes (ex: `form--taxonomy-term-form`) then there is no match while the underscore version will (`form__taxonomy_term_form`).

For more details: [drupal_find_theme_templates](https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/drupal_find_theme_templates/9.2.x) and [theme_get_suggestions](https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/theme_get_suggestions/9.2.x).

left23:
@orakili @attiks 
We've covered the taxonomy term edit form template overrides in the PR https://github.com/UN-OCHA/rwint9-site/pull/59
so I am repurposing this as the Theme Switcher module PR.
Can be merged to develop any time.
Theme Switcher config might still need tweaking as we go.

left23:
@orakili I can't seem to add you as reviewer, I guess since you opened this initially.

attiks (approved):
YOLO

PR #76, opened 2021-09-03
[RW-121] Custom fields

orakili:
Ticket: RW-121

This PR ports the custom fields used in the entity edit form.

It also adds the form customization for reports, jobs and training.


orakili:
Merging that. The country, disaster and source from customizations will be added in separate PRs.

PR #77, opened 2021-09-06
RW-54 footer partials

left23:
1. Add copyright partial with RW info, logo and styles
2. Adjust Social Media partial
4. Linting


orakili (approved):
Looks good, thanks.

PR #70, opened 2021-08-24
RW-114: field_sections

attiks:

PR #78, opened 2021-09-06
RW-127 Node edit forms

left23:
Mainly make sure all form fields are accounted for per content type, and note on the issue which ones are missing.
Some tweaks to form styles and TOC styles


orakili (approved):
@left23 Looks good thanks. Merging it as it's a bit of blocker for https://github.com/UN-OCHA/rwint9-site/pull/76.

I've added the same logic as the term forms at the bottom of the node edit templates to ensure all the fields not already rendered are rendered.

I also added a couple of new fields and changed the list of fields for the topics to match @attiks' work in #70.

There will probably be a few changes to the revision once I start on that.

Follow-up ticket to use the `reliefweb-entities-table-of-contents` template in the node/term forms: RW-131.

PR #79, opened 2021-09-06
RW-128 Term edit

left23:
Ensure all form elements are present and refine styles where possible.


orakili:
@left23 I just noticed that the form to create a term has a different route: `entity.taxonomy_term.add_form` so [this](https://github.com/UN-OCHA/rwint9-site/blob/d634e1d212892fa04dcfba540e810da5591e98d9/html/themes/custom/common_design_subtheme/common_design_subtheme.theme#L138) probably needs to be changed to check both routes.

left23:
@orakili I think this is ready for review, if we consider this to only include the template overrides and printing the relevant fields.

I added a [second route to the conditional](https://github.com/UN-OCHA/rwint9-site/pull/79/commits/95664169f00935c4cd39b74c8609cf58a04d1c09#diff-430989609a1ab8b23f88880664f47dee57398919273b2199a89fa8625fa8ea35R138) as you mentioned, but I'm not sure what to look for to test this. The `form--taxonomy-term-disaster-form.html.twig` override file seems to be being used for both an existing term edit form and a new tern edit form, regardless.

orakili (approved):
@left23 Thanks. The form suggestion from the base theme takes over when there is a template for a vocabulary. The `add_form` route check is only useful for the terms without a specific form so the `form__taxonomy_term_form` suggestion is used when creating a new term.

PR #80, opened 2021-09-13
RW-129 User pages

left23:
- Add user paths to theme switcher rule
- Add form fields to user template override file
- Add admin menu icon
- Bookmarks item styles
- RW-123 Admin menu shows for specific roles and user 1
- Fix taxonomy TOC sticky nav
- Adjust Source term field names

## To test
Import config so the User pages use the `common_design_subtheme`

The admin menu should appear to user 1, and any users with editor, administrator, webmaster roles


left23:
@orakili I think this is ready enough for review at your leisure.

html/themes/custom/common_design_subtheme/templates/cd/cd-footer/cd-footer.html.twig
```diff
@@ -18,7 +18,7 @@
       {% include '@common_design_subtheme/cd/cd-footer/cd-copyright.html.twig' %}
     {% endblock %}
 
-    {% if logged_in %}
+    {% if ('editor' in user.getroles) or ('webmaster' in user.getroles) or ('administrator' in user.getroles) or (user.id == '1')%}
```
orakili, line 21: For info, at some point, the module that handles a few things related to users will be migrated. We can add a permission for the admin menu there and remove the check on the roles.

html/themes/custom/common_design_subtheme/templates/user/form--user-form.html.twig
```diff
@@ -0,0 +1,50 @@
...
+    {{ element.account.notes }}
+  </fieldset>
+
+  <fieldset id="actions">
+    <legend class="visually-hidden">{{ 'Form submission and other actions'|t }}</legend>
+    {{ element.actions }}
+  </fieldset>
+
+  {{
+    form|without([
```
orakili, line 38: `form` is not defined. The user form is similar to the taxonomy one, so probably need the loop with the `skip_elements` rather than the `form|without`.

orakili (changes requested):
Looks good, thanks. The only thing is that the code to display the remaining fields in the `form--user-form.html.twig` need to be changed I think.

html/themes/custom/common_design_subtheme/templates/user/form--user-form.html.twig
left23: @orakili 
When I add
```

  {%
    set skip_elements = [
    'name',
    'account.current_pass',
    'account.mail',
    'account.pass',
    'account.reset_pass',
    'account.status',
    'account.roles',
    'account.field_notes',
    'actions',
  ]
  %}

  {% for key, child in element if key|first != '#' %}
    {{ key not in skip_elements ? child }}
  {% endfor %}
```

The result is duplicate fields. I changed it to use `form` for this reason but I see it's not defined, as you said.
When I don't add anything below the fields I print directly, I get the intended result i.e. a single set of fields printing in the order as they are in template markup.

left23:
I snuck a few taxonomy-related commits in

orakili (approved):
Looks good, thanks.

PR #82, opened 2021-09-16
[RW-131] Consolidate forms and use TOC template

orakili:
Ticket: RW-131

This consolidates the node, term and user forms trying to align with what the other admin themes do with nodes; and uses the `reliefweb_entities_table_of_contents` theme to reduce duplicate code.

This also adjusts the page title of the term forms including the RW profile and user posting rights ones.

The code in the `common_design_subtheme.theme` to add the `taxonomy_term_edit_form` and `user_edit_form` templates could be moved to the CD base theme: ticket: CD-340.

Also sneaked in a couple of tiny fixes for the admin menu and user dashboard.

## Test

Clear cache and check the different forms available via the "admin menu >  create" and also the user edit form. 

Links in the tables of contents should jump to the proper sections and the fields below the "save" should be limited to the revisions, url aliases, book outline and the like.


left23 (approved):
@orakili thanks, the consistency is really helpful. 
I've added one commit.

PR #85, opened 2021-09-22
RW-144: Migrate bookmarks

attiks:


orakili (dismissed):
Thanks. 

(Just tiny change to move that to the `reliefweb_extra` group as it's optional data that would be better migrated after the main content.)

orakili (approved):

PR #88, opened 2021-09-23
RW-156: Limit allowed tokens

attiks:


orakili (commented):
@attiks I cannot get this to work, the `tokenCallback` is never called and I cannot see what code would call it.

Also looking at what the token_filter module does, I feel that we could remove the module altogether and instead of extending the `TokenFilter` and basically disabling most of its features, simply have our own filter that does the disaster map token replacement. 

That would be more explicit and we would remove a dependency on a module we are not really using.

attiks:
Token filter module removed, callback fixed

html/modules/custom/reliefweb_utility/src/Plugin/Filter/ReliefwebTokenFilter.php
```diff
@@ -0,0 +1,112 @@
...
+    return $this->t('Allow a limited list of tokens and replace with their values.');
+  }
+
+  /**
+   * Token callback to limit allowed tokens.
+   */
+  public static function tokenCallback(&$replacements, $data, $options, $bubbleable_metadata) {
+    foreach ($replacements as $key => $replacement) {
+      if (strpos($key, '[disaster-map') === FALSE) {
+        unset($replacements[$key]);
```
orakili, line 107:
```suggestion
        $replacements[$key] = !empty($options['clear']) ? '' : $key;
```

If we unset the replacement, then the original key (ex: `[site:name]`) is left in the text regardless of the `clear` setting.

The above suggestion fixes that: removing the token if clear is checked and keeping the token text otherwise.

orakili (changes requested):
Thanks. See comment below for an issue in the callback.

Also we need to update the composer files to remove the `token_filter` module but that should be done after this is merged I think, to avoid a missing module issue when importing the config.

orakili (approved):

PR #89, opened 2021-09-23
RW-154

attiks:


orakili (approved):
@attiks Working well, thanks! 

I have a few fixes and changes (notably about the icons) that I will put in another PR tomorrow or so.

PR #91, opened 2021-09-24
[RW-99] Term page title

orakili:
Ticket: RW-99

The handling of the page title in the CD theme was changed in 5.0.0. This caused to page title on the taxonomy term pages to not be displayed due to some modifications to the functions handling that. This PR fixes this problem.

This also includes a tiny unrelated fix to prevent a PHP error when the database is empty which only happens when resetting the site before testing the migration and clicking the subscriptions tab.

## Testing

Simply check that the page title is present on some country, disaster and organization pages.

### DIFF
--- a/html/themes/custom/common_design_subtheme/common_design_subtheme.theme
+++ b/html/themes/custom/common_design_subtheme/common_design_subtheme.theme
@@ -181,17 +181,14 @@ function common_design_subtheme_form_user_form_alter(array &$form, FormStateInte
  * @see common_design_get_block_render_array()
  */
 function common_design_subtheme_preprocess_taxonomy_term(&$variables) {
+  $term = $variables['term'] ?? NULL;
   $view_mode = $variables['view_mode'] ?? $variables['elements']['#view_mode'] ?? '';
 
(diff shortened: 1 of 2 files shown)


html/themes/custom/common_design_subtheme/common_design_subtheme.theme
```diff
@@ -181,17 +181,14 @@ function common_design_subtheme_form_user_form_alter(array &$form, FormStateInte
...
   // Prepare the title and local tasks so we have better control over where
   // to display them for content in full term.
-  if ($view_mode === 'full') {
-    $variables['title'] = common_design_get_block_render_array('page_title_block');
-    $variables['local_tasks'] = common_design_get_block_render_array('local_tasks_block');
-    // Copy the title attributes.
-    if (!empty($variables['title']) && !empty($variables['title_attributes'])) {
-      $variables['title']['#title_attributes'] = $variables['title_attributes'];
-    }
+  // @todo review after term perview is added.
```
attiks, line 189: Typo: perview -> preview

attiks (approved):

PR #92, opened 2021-09-30
[RW-156] Remove token filter leftovers

orakili:
Ticket: RW-156

This removes some leftovers of the now unused `token_filter` module.


orakili:
@attiks The Travis build may fail due to a rate limiting issue when pulling some images from ECS (see OPS-7810) but if you don't find any problems with the changes in this PR and agree then I'll merge it regardless of the travis build status.

attiks (approved):

PR #96, opened 2021-10-05
RW-125 user forms continued

left23:
More style adjustments to user forms, to incorporate the page title on form edit pages, followed by the local tasks.
Some linting and for now, resetting the flow on the user forms where fieldsets have the border-left to prevent a white space between fieldsets.


orakili (approved):
Looks good. Thanks.

PR #100, opened 2021-10-06
RW-162: Importer

attiks:


attiks:
Terms being referenced need status 1

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,581 @@
...
+      'field_country' => $this->mapCountries((array) $data->field_country),
+      'field_how_to_apply' => (string) $data->field_how_to_apply,
+      'body' => [
+        'value' => $this->validateBody((string) $data->body),
+        'format' => 'markdown',
+      ],
+      'field_job_type' => $data->field_job_type[0] ? (array) $data->field_job_type : 0,
+      'field_job_experience' => $data->field_job_experience[0] ? (array) $data->field_job_experience : 0,
+      'field_source' => $data->field_source[0] ? (array) $data->field_source : 0,
+      'field_theme' => $data->field_theme[0] ? (array) $data->field_theme : 0,
```
orakili, line 279: Nitpicking but casting to `array` seems unnecessary here and above. If `$data->field_theme[0]`(for example) is set then `$data->field_theme` is an array.

Also, shouldn't the default be `[]` rather than `0`?

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,581 @@
...
+
+    $job = Job::create($values);
+
+    $this->validateAndSaveJob($job);
+  }
+
+  /**
+   * Create a new job.
+   */
+  protected function updateJob(Job $job, $data) {
```
orakili, line 290: Same comment as above about checking if the fields are sets (ex: `isset($data->field_career_categories[0])`) to avoid PHP notices/warnings.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,581 @@
...
+    $this->validateAndSaveJob($job);
+  }
+
+  /**
+   * Validate and save job.
+   */
+  protected function validateAndSaveJob(Job $job) {
+    $violations = $job->validate();
+    if (count($violations) === 0) {
+      $job->setNewRevision(TRUE);
```
orakili, line 315: The revision user should be set to `2` (the system user): `$job->setRevisionUserId(2)`.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,581 @@
...
+    // Check if the content contains some non encoded html tags, in which case
+    // we will assume that the text is non encoded html/markdown. For that we
+    // simply search for a closing tag '</...>'. Otherwise we decode the text.
+    if (preg_match('#</[^>]+>#', $text) !== 1) {
+    $text = html_entity_decode($text, ENT_QUOTES, 'UTF-8');
+    }
+
+    // Convert the text to HTML.
+    $text = check_markup($text, $format)->__toString();
+     */
```
orakili, line 414: Why is this disabled?

orakili (changes requested):
Thanks, looks promising. There are some leftovers from adding the unit tests I think.

Some issues:

**Field validation**

The field validation relies only the drupal entity validation however in the D7 job importer there are additional validation and specific error messages that are important to identify the problems in the feed and help the partners fix the content.

Some of the validation functions also handle particularities like for the [job experience](https://github.com/UN-OCHA/rwint-site/blob/44935f32acb0fcecf1968c7bfe2d6975b226edd4/site/profiles/reliefweb/modules/features/reliefweb_import/includes/jobs.inc#L490) to accommodate some changes that were made to the specifications.

Also [the state in which to save the job depends on the validation](https://github.com/UN-OCHA/rwint-site/blob/44935f32acb0fcecf1968c7bfe2d6975b226edd4/site/profiles/reliefweb/modules/features/reliefweb_import/includes/jobs.inc#L151-L175). Some fields are mandatory and if incorrect, the job is discarded while for some other fields not passing the validation, the job is saved as draft to give a chance to the partners to edit and correct them once imported.

Sorry, I realize, that more explanation should have been given in the ticket.

attiks:
> Some of the validation functions also handle particularities like for the [job experience](https://github.com/UN-OCHA/rwint-site/blob/44935f32acb0fcecf1968c7bfe2d6975b226edd4/site/profiles/reliefweb/modules/features/reliefweb_import/includes/jobs.inc#L490) to accommodate some changes that were made to the specifications.

```
 [info] Processing Save the Children, fetching jobs from https://batchaws.adcourier.com/services/?q=U2FsdGVkX18VqPsM1EytNPcXto4ZYZvCtcH1oznaYcTs0_-CHNWO74bNNiWV8tfyRX7DgCk6XZS0EDYu29iwQ.
 [notice] Update existing job
 [error]  Unable to process Save the Children, got 0: Validation failed in field_job_experience with message: <em class="placeholder">Job years of experience</em>: this field cannot hold more than 1 values.. 
```

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,581 @@
...
+  }
+
+  /**
+   * Process XML data.
+   */
+  protected function processXml($body) {
+    $xml = new \SimpleXMLElement($body);
+    foreach ($xml as $item) {
+      // Check if job already exist.
+      if ($this->jobExists((string) $item->link)) {
```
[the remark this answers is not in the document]
orakili: Currently it's manual work: the job team notices that the jobs from a given source are not imported or saved as draft, which means some fields didn't pass the validation or there is another problem with the feed. In that case, the create a dev ticket to investigate the logs and provide information about what's wrong.

I guess part of that could be automated with some triggers when the logs are aggregated. If you have some ideas, I'm all ears :)

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
orakili: Ok, thanks.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
orakili: Roger, thanks.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,680 @@
...
+   *   Field name.
+   * @param string $text
+   *   Field text content.
+   * @param string $format
+   *   Format to which conver the text if not already HTML.
+   *
+   * @return string
+   *   Sanitized content.
+   */
+  protected function sanitizeText($field, $text, $format = 'plain_text') {
```
orakili, line 507: "plain_text" should be changed to "markdown". D7 "markdown" format was the "plain_text" one. In D9 plain_text is really plain text and "markdown" is the real markdown. That may be the reason of the mess up in the escaping when running check_markup.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,680 @@
...
+    // we will assume that the text is non encoded html/markdown. For that we
+    // simply search for a closing tag '</...>'. Otherwise we decode the text.
+    if (preg_match('#</[^>]+>#', $text) !== 1) {
+      $text = html_entity_decode($text, ENT_QUOTES, 'UTF-8');
+    }
+
+    // Convert the text to HTML.
+    /*
+    if (function_exists('check_markup')) {
+    $text = check_markup($text, $format)->__toString();
```
orakili, line 544: Text from the feeds can be in markdown format, in which case we need to convert it HTML to do the sanitation. It's possible the escaping issue you experienced was due to the wrong text format id as it has changed in D9 (see comment above: "plain_text" => "markdown").

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,680 @@
...
+   *   Text to clean.
+   * @param array $options
+   *   Associative array with the following replacement options:
+   *   - line_breaks (boolean): replace line breaks with spaces.
+   *   - consecutive (boolean): replace consecutive whitespaces.
+   *
+   * @return string
+   *   Cleaned text.
+   */
+  protected function cleanText($text, array $options = []) {
```
orakili, line 586: For info, there is new utility helper with the same code: https://github.com/UN-OCHA/rwint9-site/blob/dea62fca917f7e9c1e16d472e4be4ee51d1b7a6c/html/modules/custom/reliefweb_utility/src/Helpers/TextHelper.php#L30

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -0,0 +1,680 @@
...
+   * Note: it's using a very basic pattern matching that may not work with
+   * broken html (missing </iframe> ending tag for example)
+   *
+   * @param string $text
+   *   Text to clean.
+   *
+   * @return string
+   *   Cleaned up text.
+   */
+  protected function stripEmbeddedContent($text) {
```
orakili, line 614: This could be moved to the TextHelper: https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_utility/src/Helpers/TextHelper.php

attiks:
> For the job closing date I added the check inside validateAndSaveJob

Now changed to FutureDateWhenPublished constraint

PR #110, opened 2021-10-07
[RW-172] Fix moderation page query for compatibility with docksal mysql

orakili:
Ticket: RW-172

This should fix the issue with doksal's mysql which has stricter settings (see [ONLY_FULL_GROUP_BY](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by)).


left23 (approved):
Thanks @orakili

PR #111, opened 2021-10-07
RW-126 Moderation pages

left23:
Includes:

- Moderation block at the top of Node and Term pages
- Styles for `/moderation/content/[type]` - Status, Totals list, Table
- User posting rights Eg. `/taxonomy/term/2007/user-posting-rights` (except the bottom margin for the page title is still missing, I will do this later as I will once again try to consolidate the vertical rhythm rules)
- User Information section on Training and Job node edit forms (now as a separate component with its own template override)
- RW-206 status icons in Moderation table and Moderation information block

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-form/rw-form.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-form/rw-form.css
@@ -256,9 +256,9 @@ form #actions {
   background: #e6ecef;
   background: rgba(230, 236, 239, 0.4);
 }
-/*form fieldset + #actions {*/
-/*  border-top: none;*/
-/*}*/
+/* form fieldset + #actions { */
+/*  border-top: none; */
+/* } */
 form #actions + * {
   border-top: none;
 }
@@ -272,9 +272,9 @@ form div.form-actions {
   /* margin: 8px -8px -8px -8px; */
   margin: 0 -8px;
 }
-/*form #actions div.form-actions {*/
-/*  margin-top: -8px;*/
-/*}*/
+/* form #actions div.form-actions { */
+/*  margin-top: -8px; */
+/* } */
 form div.form-actions button {
   margin: 8px;
 }
--- a/html/themes/custom/common_design_subtheme/components/rw-user-information/rw-user-information.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-user-information/rw-user-information.css
@@ -8,9 +8,9 @@
   display: none;
 }
 .rw-entities-form-user-information__sources ul {
+  display: inline-block;
   margin: 0;
   padding: 0;
-  display: inline-block;
   list-style-type: none;
 }
 /**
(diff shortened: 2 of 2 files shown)


html/themes/custom/common_design_subtheme/components/rw-moderation/rw-moderation.css
```diff
@@ -0,0 +1,246 @@
...
+.path-moderation form #edit-actions {
+  padding: 12px;
+  text-align: center;
+  border: 1px solid #e6ecef;
+  background: #e6ecef;
+  background: rgba(230, 236, 239, 0.4);
+}
+
+/**
+ * Double datepicker widget.
```
orakili, line 97: @left23 for info, this can be removed I think as the rules for the dual datepicker were already ported to the `rw-datepicker` component: https://github.com/UN-OCHA/rwint9-site/blob/54f75af03b3c661a0b6c4438d55017daa92a0d06/html/themes/custom/common_design_subtheme/components/rw-datepicker/rw-datepicker.css#L159:

Screen Shot 2021-10-12 at 11 21 02

html/themes/custom/common_design_subtheme/components/rw-moderation/rw-moderation--status.css
```diff
@@ -0,0 +1,138 @@
...
+  display: block;
+  padding: 10px;
+  letter-spacing: 0.5px;
+  text-transform: uppercase;
+  border-left: 10px solid;
+  font-size: 14px;
+  font-weight: bold;
+  line-height: 12px;
+}
+.rw-moderation-list .rw-moderation-list [data-status="draft"],
```
left23, line 18: Double selector

html/themes/custom/common_design_subtheme/templates/form/node-edit-form--training.html.twig
```diff
@@ -16,6 +16,7 @@
  */
 #}
 {{ attach_library('common_design_subtheme/rw-form') }}
+{{ attach_library('common_design_subtheme/rw-user-information') }}
```
orakili, line 19: Now that there is a separate component for the user information, I wonder if it would not better to create an override of `reliefweb_entities/templates/reliefweb-entities-form-user-information.html.twig` in the subtheme and attack this library there. What do you think?

html/themes/custom/common_design_subtheme/common_design_subtheme.libraries.yml
```diff
@@ -263,6 +274,13 @@ rw-user:
       components/rw-user/rw-user.css: {}
       components/rw-user/rw-user-bookmarks.css: {}
 
+rw-user-information:
+  css:
+    theme:
+      components/rw-user-information/rw-user-information.css: {}
+  dependencies:
+    - common_design_subtheme/rw-icons
```
orakili, line 282: I don't think the user information section uses icons.

PR #127, opened 2021-10-13
RW-127 Selection and Autocomplete

left23:
1. Add class `rw-selection` and `rw-autocomplete` to div so proper styles work
2. Add rule to hide select element when it has autocomplete-processed attribute
3. Add D7 some rules for autocomplete button styles and replace selectors

The result should be properly styled Autocomplete widgets on the user edit forms

Eg. Disaster `/taxonomy/term/50761/edit#countries`
Selection_053

Eg. Training `/node/3758093/edit#organization`

@orakili As we mentioned yesterday, I think the only thing missing is the `data-status` attribute on this div:
```
<div data-selection="" class="rw-selection">
  <div data-value="36436" data-index="512" data-shortname="Insight Share">...
```
which I think you said would also add the same attribute to the rw-selection list results
```
<ul class="simpleautocomplete-selector" role="listbox" id="simpleautocomplete-selector-1" hidden="">
  <li id="simpleautocomplete-selector-1-suggestion-0" role="option" tabindex="-1" class="simpleautocomplete-suggestion" aria-selected="true">...
```

Once those attributes are in place, these style should kick in:
https://github.com/UN-OCHA/rwint9-site/blob/develop/html/themes/custom/common_design_subtheme/components/rw-form/rw-form.css#L391
In D7 these styles were in `form/base.css` https://github.com/UN-OCHA/rwint-site/blob/a60043df21ca200da35ea53cf8a91f6ffa65d4e1/site/profiles/reliefweb/themes/kobe/css/form/base.css#L676 so I added them into `rw-form.css` but maybe they should be included in the `rw-autocomplete` or `rw-selection` component.


orakili:
Closing in favor of https://github.com/UN-OCHA/rwint9-site/pull/135 that adds the necessary classes in the autocomplete JS file as the styling was already present in `rw-autocomplete.css`.

PR #128, opened 2021-10-13
RW-126 Moderation table sorting

left23:
Styling only
- add rules for sortable table column icons 
- add icon files from D7
Selection_055


left23:
Yep, closing in favour of #134

PR #130, opened 2021-10-14
RW-177 Preview

left23:
- Hide view mode select element
- Style Back button so it's easily visible. Its style is similar to the "Skip to main content" link, and I need to test with proper user access to see if their placements conflict (Eg. Will the admin toolbar always be present when users can Preview?)

Selection_058

Selection_059

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-preview/rw-preview.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-preview/rw-preview.css
@@ -1,24 +1,27 @@
-/* Override Core preview rule position to add Preview label */
-.node-preview-container {
-  position: relative;
+.node-preview-container,
+.term-preview-container {
   padding: 0;
 }
-.node-preview-container header {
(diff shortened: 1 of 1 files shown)


orakili (commented):
@left23 Looks good, thanks! 2 things I noticed:

1. The preview for the terms (ex: country) is not styled. I think it needs an override of `form--term-preview-form-select.html.twig` like done for the node one.
2. The "back to content editing" is not sticky anymore. I thought it was useful when reviewing a long document. What do you think?

Also wondering if another color for the "preview" banner would not make it more obvious.

left23:
@orakili Yes, good point about the sticky header. I've made it sticky now and added top padding to the `dialog-off-canvas-main-canvas` div so everything is still visible before the page is scrolled.

For now, I've made the header red `--cd-reliefweb-dark-red: #eb1405;` Let me know if you think another colour is better.

I've also added another template override for the term preview.
Selection_067

PR #136, opened 2021-10-14
Test Localization helper

attiks:


html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php
```diff
@@ -113,7 +113,6 @@ public static function collatedKsort(array &$items, $language = NULL) {
         $items = $reordered;
         return TRUE;
       }
-      return FALSE;
```
orakili, line 116: This changes the logic and makes it less consistent with the other methods that return FALSE when collated sorting fails.

Defaulting to ksort when the collated sorting of the keys fails may not be appropriate. I think it's better to restore the `return FALSE`.

html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php
```diff
@@ -134,10 +133,13 @@ public static function getCollator($language = NULL) {
     $language = static::getLanguage($language);
 
     if (!isset($collators[$language])) {
-      if (function_exists('collator_create')) {
-        $collator = collator_create($language);
+      $collator = static::collatorCreate($language);
 
-        switch (intl_get_error_code()) {
+      if (!$collator) {
+        $collators[$language] = FALSE;
```
orakili, line 139:
```suggestion
        $collator = FALSE;
```

To be compatible with line 166 below: `$collators[$language] = $collator;` otherwise this will store a `NULL` if collator_create failed in `static::collatorCreate`.

Alternatively, have `static::collatorCreate` return FALSE if the collator couldn't be created and replace the the `if/else` with a `if ($collator !== FALSE) {` before the switch.

html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php
```diff
@@ -159,17 +161,35 @@ public static function getCollator($language = NULL) {
...
   }
 
+  /**
+   * Get the collator for the given language.
+   *
+   * @param string $language
+   *   Language for which to return a Collator. Defaults to the current
+   *   language.
+   *
+   * @return \Collator
```
orakili, line 179:
```suggestion
   * @return \Collator|null
```

html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php
```diff
@@ -159,17 +161,35 @@ public static function getCollator($language = NULL) {
...
+   *   Collator.
+   */
+  protected static function collatorCreate($language) {
+    return collator_create($language);
```
orakili, line 183: Maybe return FALSE when collator_create fails (see other comment above).

orakili (changes requested):
Looks good. (Just needs a change to the getCollator).

However, I feel that the tests don't really test the collated sorting as it's only testing `a`, `b`, `c`.

I think it would be good to add tests with strings with accents etc.

attiks:
Will add more tests with accented characters

html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php
```diff
@@ -134,42 +136,62 @@ public static function getCollator($language = NULL) {
...
-          default:
-            // Some other error happened, we mark the collator as FALSE so that
-            // the collated_(k)sort functions can default to the basic (k)sort.
-            $collator = FALSE;
-        }
+      $collator = static::collatorCreate($language);
 
-        $collators[$language] = $collator;
+      if (!$collator) {
+        return FALSE;
```
orakili, line 142:
```suggestion
        $collators[$language] = FALSE;
        return FALSE;
```

In my opinion `$collators[$language]` should be set to `FALSE` in case of failure to avoid calling `collatorCreate` again.

PR #135, opened 2021-10-14
[RW-127] Autocomplete

orakili:
Ticket: RW-127

This is a replacement of https://github.com/UN-OCHA/rwint9-site/pull/127 that updates the autocomplete widget to use the existing classes for styling.

**Note**: this also changes the status attribute (data-moderation-status) to be compatible with what the Drupal field widget adds (see https://github.com/UN-OCHA/rwint9-site/pull/131). The styling will be updated in another PR.

PR #137, opened 2021-10-14
[RW-127] Moderation status

orakili:
Ticket: RW-127

**Note** this is against the branch of the autocomplete PR: https://github.com/UN-OCHA/rwint9-site/pull/135

This consolidates the different places where the moderation status is used:

- Allow to add individual attributes to `radios` and `checkboxes` elements (so that we can add the `data-moderation-status` to the filters on the moderation pages).
- Create new components: `rw-revision-message`, `rw-moderation-status` and `rw-moderation-information` for more granularity (that will be useful for the revisions).
- Reduce the moderation status selectors to only target the `data-moderation-status` attribute
- Add a few rules to the `rw-autocomplete` and `rw-selection` to adjust the display for elements with the `data-moderation-status` attribute.
- Revisit the libraries dependencies (for example so that the `rw-selection` is loaded when using the `rw-autocomplete` as it's kind of set)

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-revision-message/rw-revision-message.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-revision-message/rw-revision-message.css
@@ -17,6 +17,3 @@
 .rw-revision-author {
   font-style: italic;
 }
-.rw-moderation-information .rw-revision-message {
-  font-size: 16px;
-}


orakili:
@left23 Yes, it's a new addition and only relevant for the reports. This should help remind the editors to update a country's profile when creating a report if the tagged country is red (ongoing humanitarian situation). Sorry I didn't mention it.

This is controlled through the UI (field settings).

PR #138, opened 2021-10-14
[RW-129] Country subscriptions autocomplete

orakili:
Ticket: RW-129

This refactors a bit the way the form "enhancements" are applied so the country subscriptions select is transformed into an autocomplete widget like on D7.

This should also make it easier to do the same on other forms (like the contact one when ported).


left23 (approved):
Approving. I can see the autocomplete widget. It's not styles (the select isn't hidden) but I think that will come from #135

PR #140, opened 2021-10-14
RW-129 subscription updates

left23:
Follow up from https://github.com/UN-OCHA/rwint9-site/pull/138
I noticed the select element was higher than the others.
`normalize.css` used `box-sizing: content-box` for `input[type=search]`
Adding cd-form to the form template overrides this with `border-box`

Also ran `npm run sass:lint-fix` and this caught the recent guidelines stylesheet.


left23:
@orakili I've just noticed I can do a bit more on this. No need to review yet.
Selection_065

PR #141, opened 2021-10-14
RW-132 Subscription updates based on Litmus testing

left23:
1. recommended by litmus to add `lang` attribute to html tag so screen readers do not default to user language


left23:
@orakili I anticipated adding the work for the plain text emails (and maybe others) to this PR but since RW-116 is pending, let's merge this on its own.

PR #146, opened 2021-10-15
[RW-160] Datepicker

orakili:
Ticker: RW-160

This changes the datepicker widget so it can apply to date elements and use the existing styling.

Screen Shot 2021-10-15 at 20 22 39


left23 (approved):
@orakili Thanks, looks good.
I've added a few styling rules to this PR.
I noticed there's a Publication date field on the D9 Blog node that isn't using the widget, but this doesn't appear on the D7 staging site.

PR #148, opened 2021-10-18
Test filters

attiks:


orakili (dismissed):
Looks good to me. 

It may be missing 2 other tests:

- `[disaster-map:XX-YY]`: map with 2 types of disasters (ex: floods and flash floods)
- `[disaster-map:123456-123457-123458]`: map with fixed disaster ids

PR #149, opened 2021-10-18
RW-125 user forms

left23:
Fine tuning of User editor form


left23:
@orakili This is good for review when you're ready. 

There's still a lot of commented-out rules and I've started moving generic form-related styles to the base theme so there is also unnecessary duplication with rules that override the same rules in the base theme.

Drupal field styling can be hard because of all the nesting possibilities, so I'd like to merge this and then remove irrelevant rules in another step that will be a bit more controlled.

PR #153, opened 2021-10-20
Rw 205

attiks:


html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -576,6 +576,27 @@ protected function validateHowToApply($data) {
...
 
+  /**
+   * Validate city field.
+   *
+   * @param string $data
+   *   Raw data from XML.
+   */
+  protected function validateCity($data) {
+    // Clean the field.
+    $field_city = $this->sanitizeText('field_city', $data, 'markdown');
```
orakili, line 587: @attiks The city field only accepts a plain text string. Sanitation should probably do something like that instead to be closer to the D7 validation:

```
$field_city = TextHelper::cleanText(strip_tags($data), [
  'line_breaks' => TRUE, 
  'consecutive' => TRUE,
]);
```

attiks: Changed, thanks

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -597,6 +597,34 @@ protected function validateCity($data) {
...
+    $length = mb_strlen($field_job_closing_date);
+    if ($length !== 0 && $length !== 10) {
+      throw new ReliefwebImportExceptionSoftViolation(strtr('Invalid data for field_job_closing_date, @length characters found, format has to be yyyy-mm-dd', [
+        '@length' => $length,
+      ]));
+    }
+
+    // Make sure field can be converted to a date.
+    if ($length === 10 && !strtotime($field_job_closing_date)) {
+      throw new ReliefwebImportExceptionSoftViolation(strtr('Invalid data for field_job_closing_date, @data has to in format yyyy-mm-dd', [
```
orakili, line 623: > @data has to in format yyyy-mm-dd

Typo.

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
```diff
@@ -597,6 +597,34 @@ protected function validateCity($data) {
...
+    // Ensure the field size is reasonable.
+    $length = mb_strlen($field_job_closing_date);
+    if ($length !== 0 && $length !== 10) {
+      throw new ReliefwebImportExceptionSoftViolation(strtr('Invalid data for field_job_closing_date, @length characters found, format has to be yyyy-mm-dd', [
+        '@length' => $length,
+      ]));
+    }
+
+    // Make sure field can be converted to a date.
+    if ($length === 10 && !strtotime($field_job_closing_date)) {
```
orakili, line 622:
```suggestion
    if ($length === 10 && !date_create_from_format('Y-m-d', $field_job_closing_date)) {
```

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
attiks: It's meant to be `data`, since it's the user input

html/modules/custom/reliefweb_import/src/Command/ReliefwebImportCommand.php
attiks: Thanks

PR #156, opened 2021-10-21
[OPS-7850] add and configure config_split

lazysoundsystem:
https://humanitarian.atlassian.net/browse/OPS-7850

As with the others, making local dev config_split consistent.

Note, this renames config-dev directory to config_dev (we had both the hyphen and the underscore in different repos, more with the underscore).

Almost no change to the configuration itself - but does remove config-dev/core.entity_view_mode.config_split.token.yml as I wasn't sure what it's for - if this is useful, I can restore it.


left23 (dismissed):
This seems to work now @lazysoundsystem, thanks for the additional info.

orakili (approved):
Working well.

@lazysoundsystem just for info, I found that `var_dumper` for `dpm` is much faster and memory efficient than `kint`.

PR #163, opened 2021-10-25
RW-31 Frontend QA

left23:
- Style tweaks based on comparing D7 and D9, and browser and device testing. See screenshots logged on RW-31
- Sets up Help menu and menu block in header action area. Run `drush cim`. Add `/help' menu item manually.
- Includes refactoring for Sticky nav Table of Contents RW-217 which involved changes to display values (flexbox), and at what breakpoints they change -  VRT and user form (logged in) testing to verify no negative effects on the layout.
- Relies on `cd-form` component updates to base theme https://github.com/UN-OCHA/common_design/pull/254
 Run `composer update unocha/common_design`
- Moves many functions from the subtheme to the base theme's `common_design.theme` file
- Moves Tabs block above Page title block in Content region for Form edit pages and User pages and adjusts page title styles so underline is in the right place
- Set full image path in `rw-icons` component to fix bug in Safari to do with relative paths for background images and CSS custom properties - see https://stackoverflow.com/questions/66752308/css-background-image-relative-path-var-on-safari-not-loading-image


left23:
@orakili I've deployed this branch via #165 to https://dev.reliefweb-int.ahconu.org
Please review at your leisure. I've been running visual regression tests and doing a lot of manual testing and will continue to do so.
In following PRs I will start to remove commented out code and tidy up, once I am happy with how the base theme branch (also ready for review soon) is working.

orakili (approved):
@left23 Thanks! I've checked all the pages and they looked good.

I've added a few commits:

- Updated the `rw-icons` script to use the full path when generating the css files
- Fixed the spacing for the pager in river with columns
- Removed unnecessary border and padding for pager on moderation page
- Added the country slug styling to training pages (in addition to reports and jobs) with some simplifications
- Fixed some spacing issues on river pages (notably between the search results and letter navigation. I think we could use the `cd-flow` approach there.
- Fixed and simplified the styling of the page title across pages. On the D7 sites, there are inconsistencies regarding the bottom padding before the red border on the Org. and Blog rivers so I consolidated to 12px here.

PR #166, opened 2021-10-28
RW-31 update tests from base theme and adjust for RW

left23:
Run `nvm use` then `npm run e2e` in sub theme
The `MainMenuDropdown` will fail as RW does not have second level menu item in the main menu.


left23:
@orakili These tests are directly from the most recent base theme. I've adjusted only the URLs and menu item titles for the OCHA Services Menu and Footer menu.

orakili (approved):
Working well when testing against the dev site.

PR #167, opened 2021-11-02
RW-243 user posting rights and RW-244 image alt text display

left23:
RW-243
1. Set font-size for User posting rights ul to match D7
2. Vertical rhythm for User posting rights form following page title
3. Fix typo in page title

RW-244
1. Set display and overflow to control alt text display


orakili (approved):
Thanks! And sorry for the typo...

PR #170, opened 2021-11-03
RW-31 more frontend QA

left23:
Based on recent QA and discussion
-  RW-100 style checked/uncheck revision like D7
Selection_150

- Hide country slug in all sections on Country pages

- Change pagination heading from h4 to h3
- On pages with `rw-river-page--with-advanced-search` change pagination width to span full width of container, despite its parent section width being restricted

**To Test:**
Pagination on `/organizations` should display as usual
On pages like `/updates` pagination should span full width and remain centered from 768px upwards

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-country-slug/rw-country-slug.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-country-slug/rw-country-slug.css
@@ -38,3 +38,8 @@
 .rw-entity-country-slug a:active {
   text-decoration: underline;
 }
+
+/* Hide country slug on Country pages */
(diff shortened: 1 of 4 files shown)


left23:
@orakili This can be reviewed at your leisure.
I've changed the pager nav heading to h3 and I think that corresponds with the D7 site if we consider the change in heading levels throughout the D9 site.
For https://stage.reliefweb-int.ahconu.org/updates the river article titles and the pager title were h4, now they are h3
For https://stage.reliefweb-int.ahconu.org/organizations the article titles and the pager title were h4, now they are h3

orakili (approved):
Thanks!

PR #171, opened 2021-11-04
[RW-124] Headlines widget

orakili:
Ticket: RW-124

This ports the headlines widget to select the headlines on the homepage.

## CSS component changes

- The `rw-headlines--admin` component was removed as the JS and associated CSS for the headlines widget are now in the `reliefweb_homepage` module. So it's replaced by the `reliefweb_homepage/headlines-widget` component that probably needs to be extended in the subtheme to replace the colors with the CSS variables.
- The `rw-article` component was updated (last commit) to use class selectors instead of the element selectors (ex: `.rw-river-article header` --> `.rw-river-article .rw-river-article__header`. The element selectors for footers were impacting the thumbnail's footer for example instead of just applying the rules to the article's footer.

## Tests

`drush cim` to import the permission.

1. Ensure the API is pointing at the stage one, not prod to have more synchronized data and see the headline images.
2. Log in as an Editor
3. Go to the homepage, click "edit" to show the widget (headlines already in use have an opaque overlay).
4. Drag and drop headlines from the widget to replace articles from the homepage's headlines section (duplicates are highlighted in red).
5. Save, then reload the homepage and confirm it's showing the new selection.

Also confirm that the "edit" button doesn't appear when visiting the page as anonymous or normal authenticated user.


orakili:
@left23 I tested that the changes (element selectors to class selectors) made to the `rw-article` component (https://github.com/UN-OCHA/rwint9-site/pull/171/commits/a3cab85ab03d2b59826e00920cb62502fac2064c):  didn't impact negatively the different rivers across the site and they all look unaffected. However if you feel those changes don't belong here, please feel free to revert the commit.

left23 (approved):
@orakili Great, thanks. I tested as you described and it all works well.

left23:
@orakili Those changes to the selectors make sense. Merging this.

PR #172, opened 2021-11-04
[RW-253] Fix active link in book navigation

orakili:
Ticket: RW-253

This fixes the logic to determine the active link in the book navigation:

Screen Shot 2021-11-04 at 18 08 21

## Tests

Simply check that the highlighted entry in the book navigation corresponds to the current page.


left23 (approved):
Thanks, approving.

PR #173, opened 2021-11-04
[RW-252] Prevent display of text section with 'empty' content

orakili:
Ticket: RW-252

This prevents the display of sections on country/disaster/topic etc. pages when their content is basically empty (ex: /topic/humanitarian-financing overview is just 2 blank lines so it shouldn't be displayed).


left23 (approved):
Thanks, approving.

PR #174, opened 2021-11-04
RW-37 Replace variables

left23:
- Refactor colour variable names Eg. add keyword `brand` and shade `--light` as BEM-like modifier to differentiate between regular colours (used in moderation labels)
- Reset CD Sass variables to use OCHA brand colours, and instead use RW brand colour variables for `implementation` values, for clarity.
- Replace colour values throughout
- extend custom reliefweb modules and add components for stylesheet overrides with the colour values as CSS custom properties
- As discussed on RW-82 `--cd-reliefweb-orange--alert` might be replaced with different orange/yellow colour


left23:
@orakili I've done extensive visual regression tests between my local and the dev site with 'develop' branch deployed to check that none of the colour changes result in unintended changes. IMO all good. Please review at your leisure.

orakili (approved):
@left23 Looks really good, thanks.

The only thing is that, in my opinion, the rules for the user posting rights **form** and the rules for to style the elements with the `rw-user-posting-right` class should be in separate components.

The reasoning is that elements with the `rw-user-posting-right` class may appear on many different pages (revisions block on entity forms and node pages, moderation pages etc.) on which the form rules are not relevant.

I think that's my mistake because I named the component `rw-user-posting-rights` while it should be `rw-user-posting-right` to avoid confusion with the posting rights form.

I created RW-257 for that.

Merging.

PR #180, opened 2021-11-10
RW-257 User posting right/s

left23:
RW-257
- split libraries for user posting right/s
- add as dependency where relevant

I've checked 
1. `node/3758093/edit` (Training user information)
2. `taxonomy/term/2007/user-posting-rights` Organization user posting rights
3. `taxonomy/term/2007/edit#edit-revision-information` or `taxonomy/term/2007/user-posting-rights` Organization revisions user posting rights
4. `moderation/content/training`

I don't see the class `rw-user-posting-right` applied to the revision history spans. @orakili Can you take a look? And also let me know if there are other places this component appears that I neglected to check.
Selection_155

### DIFF
--- a/config/theme_switcher.rule.request_path.yml
+++ b/config/theme_switcher.rule.request_path.yml
@@ -12,5 +12,5 @@ admin_theme: common_design_subtheme
 visibility:
   request_path:
     id: request_path
-    pages: "/moderation/content/*\r\n/user/*/edit\r\n/user/*/notifications\r\n/user/*/bookmarks\r\n/user/*/posts\r\n/user/*/contact\r\n/admin/community-topics"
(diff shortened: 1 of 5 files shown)


orakili (approved):
@left23 Thank you and nice catch for the revisions.

What happened is that the revision block's content was not rendered with the subtheme (loaded via JS so not in the page context) and so the subtheme templates (notably the `reliefweb-moderation-user-posting-right.html.twig` one) were not used (and thus the library was not attached).

I've added the path to get the revision history in the theme switcher settings and it's now properly rendered using the subtheme:
Screen Shot 2021-11-11 at 16 25 10

Something to keep in mind if we find something weird with the styling of something rendered asynchronously. Thanks again.

PR #181, opened 2021-11-11
[RW-25] File attachments

orakili:
Tickets: RW-25, RW-120

This PR adds the docstore integration for the report attachments.

Note: this also adds a `_is_migrating` switch to prevent executing some of the `hook_entity_*()` like queueing notifications (see RW-120).

## Concepts

See [README](https://github.com/UN-OCHA/rwint9-site/blob/RW-25/html/modules/custom/reliefweb_docstore/README.md).

## Styling

The frontend styling was already handled by @left23 in https://github.com/UN-OCHA/rwint9-site/pull/185 and that works well.

The backend styling (file widget styling) needs some work to align the subfields nicely and to style the "Actions" (delete and replace).

There is also a problem with the field item re-ordering: the tabledrag script doesn't process the table properly. It will be tackled in RW-269 and should hopefully not require extra styling.

## Testing

The focus of the testing on this PR is file uploads and downloads.

The `reliefweb_dosctore` module has 2 possible modes: local storage (default setting) or storage in the docstore.

If you have a **local docstore** instance, please use to test:
- download a fresh backup of the docstore DB or create a "reliefweb" user, get the user api key and UUID.
- make sure your local docstore instance is reachable by your local RW9 site
- add the following in your `settings.php`:

```php
$config['reliefweb_docstore.settings']['docstore_url'] = 'http://url.of.your.local.docstore';
$config['reliefweb_docstore.settings']['api_key'] = 'the.reliefweb.user.api.key';
$config['reliefweb_docstore.settings']['local'] = FALSE;
$config['reliefweb_docstore.settings']['provider_uuid'] = 'the.reliefweb.user.uuid';
```

### Preview generation requirement

The preview generation is done via the [mupdf](https://mupdf.com/) tool, in particular its utility tool: `mutool`.

if you use one of the unocha php docker images, then you're good to go. Otherwise you somehow needs to install it so it can be executed by the RW9 drupal site. The default location for the tool is `/usr/bin/mutool` but you tell Drupal to look elsewhere via a `drush sset mutool /another/location/for/mutool`.

If you cannot install it, then the file uploads and downloads should still work but you will obviously not see the previews and that "map" and "interactive" reports will not display useful information.

### Upload testing

Create a report node. Fill in the title, body and other required fields. Then go the "Files" section in the form to upload attachments (click "choose file(s)" then "upload file(s)).

For testing PDF and checking the preview is properly generated, you can go to https://reliefweb.int/updates?search=_exists_%3Afile and download a PDF from one of the reports to compare with the D9 upload.

- Try uploading a file with an extension that is not in the list, check that there is a valid error message.
- Try uploading a file with a size superior to the limit (you can change the limit on `/admin/structure/types/manage/report/fields/node.report.field_file` if you want), check that there is a valid error message.
- Try uploading a single file, check that the file is added to the list.
- Try uploading several files at once and check that they all appear in the list.
- Try uploading a non PDF file, check that there are only 3 columns (file name (+ description + language); actions; weight)
- Try uploading a PDF file and verify that a preview is generated.
- Try changing the rotation and preview page and check that the preview corresponds to your choices.
- Try setting the preview page to "no preview" and confirm that there is no preview...

Save the report and confirm that the uploaded files appear on the report node page (download testing and expectations are below).

***Docstore***

You should see a new file under `/admin/content/media` in the docstore (or check via the API) and a new `reliefweb_document` on the last page of `https://docstore.test/admin/content` (or check via the API).

Click the media file and then the "usage" tab and check that there is one item matching the `reliefweb_document` above (or check via the API file usage endpoint).

### Download testing

**Expectations**

- attachments of **published** reports should be publicly available (check as anonymous in an incognito window for example)
- attachments of **unpublished** content should only be accessible to people with the Editor role (test as anonymous, normal logged in user and editor)

**Normal download**

- Create a report with an attachment as above and save as "published", check that the file is acccessible whether you are logged in or not.
- Create a report with an attachment as above and save as "on-hold", copy the attachment link and try access it when logged in as Editor: file should download then logged in as a normal user and not logged in: should be a 403

**Non existing file**

- Copy the file URL of the public file above and change a character from the the uuid (ex: 2 --> 3), try the URL in a browser and you should see a simple "404 Not Found" message.
- Do the same with the private file above and you should see the Drupal 403 page.

**Docstore**

Check the access log of the docstore instance for the requests to the file aboves.

### Public/Private switch

- Create a report with an attachment and save as "published", check the attachment is publicly available
- Edit the report, save as "on-hold", check that the attachment is now private
- Edit the report and save as "published", check that the attachment is public again.

**Docstore**

Check that the `private` property of the file resource changes when following the steps above.

### Full file deletion

- Create a report and upload a file as above
- Run `drush sql-query "SELECT fid, filename, uuid, uri FROM file_managed ORDER BY fid DESC LIMIT 1\G"`
- Delete the report
- Run the same command and check that the entry is gone

**Docstore**

In addition to the above, check that the `media` and the `reliefweb_document` resources are gone after deleting.

### File replacement

- Create a report with an attachment and save as "published", note the URL of the attachment.
-  Edit the report to try to replace the file by clicking "actions" and then selecting and uploading a file replacement:
   - Try uploading a file with a different extension, you should get an error
   - Try uploading a file with the same extension, you should see the "size" next to the filename change to reflect the size of the new file. If it's a PDF, a new preview should be generated.
- Save the report and check that the attachment URL is the same as before but that you get the new version of the file when downloading it

**Dosctore**

Check the revisions endpoint for the file, there should be 2 entries.

### Revisions

**Reverting to a replaced file**

- Replace a file as described above.
- Go to the "revisions" tab and revert to the revision corresponding to the first attached file
- Check that the URL hasn't changed but the file to download is the first version.

**Reverting to a deleted file**

- Create a report with an attachment and save (note the URL)
- Edit and delete the file (under "actions) and save the report, there should be no attachment anymore
- Try the file URL, you should get a 404
- Go to the "revisions" tab and revert to the revision with the file
- The file should be back with the same URL and the file should be downloadable.

**Deleting a version**

- Create a report with an attachment
- Run `drush sql-query "SELECT fid, filename, uuid, uri FROM file_managed ORDER BY fid DESC LIMIT 1\G"`, the first row is the attached file.
- Edit the report, delete the file and save
- Run the same command as above, the file entry should still be there but the `uri` should start with `private://` and contain the `uuid`
- Go the "revisions" tab and delete the first revision (the one with the attachment)
- Run the same command as above, there entry should be different as the file is gone

**Docstore**

In addition to the file record being removed from the RW9 database, the corresponding file resource in the docstore should be gone to.

---

## TODOs

- format the revisions for the files - RW-266
- check if replaced files (when migrating the delta during the production switch) are handled properly - RW-267
- keep local files on failure to upload to the docstore - RW-268
- table drag to re-order the files - RW-269


orakili:
@attiks @left23 This is ready for testing (check branch + drush cim). Please see the `testing` section in the PR description.

There are extra setup and testing instructions if you have a local docstore instance to test the integration. If you have the opportunity please test with that to ensure the RW - docstore integration works as expected.

orakili:
@cafuego If you feel like having a look, here are the rules that were added to handle both the attachment redirections (public files -> pass request to docstore), and the legacy URLs redirections (D7 file and image URLS to D9 uuid-based URLs): https://github.com/UN-OCHA/rwint9-site/pull/181/files?authenticity_token=SHeiWzFEQ91pdErdCAmkCQBVVw6SbDXA5tOyo9ZwRV9O1DEEgF2c9KJxo0WUB2buhD91bf6DeX2FKeDZp%2FHc%2BQ%3D%3D&file-filters%5B%5D=.conf&file-filters%5B%5D=.lua&file-filters%5B%5D=No+extension.

Let me know if you have opinions on this.

cafuego:
Well that escalated quickly!

left23 (approved):
I tested what I could locally, without the stack and without mupdf, and without a local docstore.

I tested the PDFs on the Dev site.

### Upload testing locally
:heavy_check_mark:  valid error when uploading file type not included in list

:heavy_check_mark: valid error when the file size exceeds limit.

:heavy_check_mark: can upload a single file and it is added to the list

:heavy_check_mark: can upload several files and they are added to the list, but each file upload is a separate action i.e. choose file, click upload, repeat, save node = multiple files are uploaded.

:heavy_check_mark: non-PDF files have 3 columns

:heavy_check_mark: Node saves and attachment appears on Report page

### Upload testing on Dev site
:x:  When the file size exceeds limit, I don't get an error. The spinner spins, and eventually stops. The file name remains in the file input and cannot be cleared so it means one must go back in the browser or click View (or otherwise travel away from the node), potentially losing other edits. I can't see a way to remove the too-large file and keep editing to save. Without clearing the file input, the browser returns a 413 Entity too large on save.

:heavy_check_mark: PDF generates a preview

:heavy_check_mark: PDF generates an accurate preview based on the page I select, and rotates correctly.

:heavy_check_mark: No preview works

### Download testing locally
:heavy_check_mark: Attachments of published reports are available to download for all users

:heavy_check_mark: Attachments of unpublished reports are not available to download except for Editors

:bangbang: Draft and On-hold nodes are visible to Auth users (but without file attachments) [make ticket]

:heavy_check_mark: On-hold file attachments return a 403 for auth and anon users. File downloads for Editors

:heavy_check_mark: Non-existing public files return a 404 in all cases

:heavy_check_mark: Non-existing private files return a 403 in all cases

### Public/Private switch locally
:heavy_check_mark: Publishing a node result in a file that is publicly available

:heavy_check_mark: Changing the node to on-hold makes the file private

:heavy_check_mark: Changing back to publish makes the file publicly available again

### Full file deletion locally
:heavy_check_mark: File is gone after deleting the node

### File replacement locally
:heavy_check_mark: Replacing the file with a different file type results in an error

:heavy_check_mark: Replacing the file with a different file size updates the file size displayed

:heavy_check_mark: The attachment URL is the same despite the file having been replaced

### File replacement on Dev site
:heavy_check_mark: When replacing a PDF, the file size and preview is updated

### Revisions locally
:heavy_check_mark: Reverting to a replaced file works as described.

:heavy_check_mark: Reverting to a deleted file works as described

:heavy_check_mark: Deleting a version works as described

### Comments
This is great @orakili Thanks for the detailed HTO

*The only comment I have from a UX point of view is that I found it a bit confusing when file name stays the same after replacing a file. I understand this makes sense to keep the URL the same and my assumption is that this is by design. 
Depending on the Editor's understanding, I think making a clear distinction between when a file is replaced versus when a file is deleted and re-uploaded might be helpful. Maybe this is something Editors are already familiar with. I'm wondering about the scenario where they might want to replace a `document_v1.pdf` with `document_v2.pdf` and whether the expectation is for the file name to reflect this.

orakili:
@left23 I've merged the styling work from https://github.com/UN-OCHA/rwint9-site/pull/199 in that branch and deployed it to the dev site. Could you have a quick look to make sure I didn't break anything and re-approve this PR if it looks ok to you?

left23 (approved):

cafuego (approved):
I am rubberstamp-approving the changes to NGINX. As long as they work, let's move them. We can refactor later if needed and nobody needs to be blocked by my time off.

PR #185, opened 2021-11-17
RW-25 Report attachments

left23:
- Update selectors from D7 to reflect BEM naming for D9 including `rw-attachment--` as class variants
- Add wrapper div inside existing `.rw-article__content` so the content body and the `.rw-attachment--report` respects their float rules.
- TEMP commented-out markup directly into Report node template override to test

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-report/rw-report.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-report/rw-report.css
@@ -4,85 +4,86 @@
  * Individual report pages. Those are made of the main content body with
  * eventual attachments.
  */
-/* Lists in content */
-.node--report--full li {
-  margin: 8px 0 0;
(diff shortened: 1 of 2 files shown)


left23:
@orakili I've left the markup I used for testing in the template override file. I've changed some class names where I thought was relevant and I thought this was the best way to communicate it. Feel free to change to fit what you are doing.

When you approve this, you can remove the commented-out markup, or I will before I merge.

orakili (commented):
Looks good. I'm keeping that open for now. I'm going to work on the field formatter tomorrow and that will be a good time to ensure the compatibility with the changes here.

orakili (approved):
Merging. I'll remove the commented markup in [RW-25](https://github.com/UN-OCHA/rwint9-site/pull/181).

PR #187, opened 2021-11-24
RW-169 Topic tokens

left23:
In D7, this used custom JS to expand and collapse the table.

For simplicity, we're opting to use the `details` and `summary` elements with its native toggle state for D9.

Selection_169

The `code` element style rule was missing so that has been added now too.


left23:
@orakili Please note the [hardcoded markup](https://github.com/UN-OCHA/rwint9-site/pull/187/files#diff-1d9be371410630a1cadb699e7fea30360c06d56db5b17175231d79bbf51f4058R60) currently present for testing should be replaced with the real thing. Let me know if I can do more to help.

orakili:
@left23 I've committed some code to generate the disaster map token help and removed the hardcoded one. The table ends up inheriting the `cd-table` style which looks ok.

I noticed that depending on the browser, the `<summary>` shows up with or without an arrow icon. For example on Chromium based browsers on Mac the arrow doesn't seem to appear:

Screen Shot 2021-11-25 at 16 23 56

While it's ok on safari or firefox:

Screen Shot 2021-11-25 at 16 39 19

According to https://stackoverflow.com/a/56649741, the default arrow can be removed via a couple of rules. It could be then replaced by the classic arrow icon. What do you think?

left23:
@orakili Thanks. I've replaced the icon with a custom one.
Screenshot from Chrome:
Selection_170
I _think_ the only other place we used the `summary` element is in the revisions, which already has its own toggle button treatment. Let me know if there are other places `summary` is used and I'll need to move these rules somewhere more generic.

orakili:
@left23 Thanks. There will be a summary element for the files to show/hide the "delete" and "upload replacement" actions. But for this I'm not sure what style is good. Merging this PR in the mean time.

orakili (approved):

PR #189, opened 2021-11-29
RW-263 Full text description remove italics

left23:
Template override to add class `description` to field description div to use generic form styling and thus remove specific rule that conflicts with the main text areas.


orakili (approved):
Thanks!

PR #194, opened 2021-12-06
Guidelines frontend additions - take 2

left23:
**Sub theme**
- extend custom module libraries for CSS
- add rules copied from custom module and adjust selectors for better BEM naming
- adjust styles for button placement
- remove wrapper div in template override in favour of single wrapper div added with JS
- override custom module JS with sub theme component JS
- adjust selector names for divs created in JS, with `rw-` prefix and BEM naming
- adjust JS so button placement matches that of D7 in the 2 cases (label, legend) - button inside legend element (in D9 button is also inside the span that wraps the legend text due to its `form-required` class), button follows label element
- change h4 tag to label for datetime field for consistency in icon button placement

**Custom module**
- remove unnecessary parent selectors in CSS, which prevents easy overriding in sub theme due to specificity issues
- Set colour of h1 explicitly so it always displays on the dark background of the aside element regardless of the generic h1 colour


left23:
@attiks @orakili
This is ready for review. I've left some code commented out, so once I get feedback, I will need a few more commits to tidy up.

orakili (commented):
@left23 I haven't fully reviewed this but, if my understanding is correct, the `guidelines` module is intended to become a contrib module so the `guidelines/components/guidelines-json/guidelines-json.css` and `guidelines/components/guidelines-json/guidelines-json.js` should probably not contain `rw-*` classes.

Meaning, those 2 files should probably be copied and overridden in the subtheme to add what is necessary to satisfy RW needs.

attiks (approved):

orakili (approved):
Looks good, thanks.

@left23 However the change from `<h4>` to `<label>` means this rule: https://github.com/UN-OCHA/rwint9-site/blob/7d04f92e4261bd5f017e689ddac3a6dbddb5f855/html/themes/custom/common_design_subtheme/components/rw-form/rw-form.css#L182-L185 doesn't apply anymore and that affects the training form:
Screen Shot 2021-12-08 at 18 15 35

PR #198, opened 2021-12-09
[RW-159] Rich editor

orakili:
Ticket: RW-159

This enables CKEditor for the job/training body and how to apply/register fields with a custom field widget to convert from markdown to HTML (what CKEditor works with) and convert back to markdown (legacy content format in RW) when saving. 

The default CKeditor styling is good enough I think:

Screen Shot 2021-12-09 at 18 57 04

We could inject the subtheme styles.css to be more like the end result, @left23 what do you think?

## Testing

- Create a job/training
- Add some content to the body for example with some styles (bold, heading, list etc.)
- Save and check that the output seems coherent with what was entered.
- Edit the node, save again, check nothing changes (there should be no new entry in the revision history at the bottom of the page as nothing changed)
- Edit the node, copy and paste some content from various sources and check that what is displayed in the editor seems correct (only the following tags are preserved: `<p> <a> <em> <strong> <ul> <ol> <li> <h2> <h3> <h4> <h5> <h6>`).

## Issue

- RW-273: pressing enter when the modal to add a link is open redirects to the link form instead of closing the modal and updating the link. Also pressing tab never focus the modal button.


left23 (commented):
When I copy the content from https://web.brand.unocha.org/demo#cd-typography into the WYSIWYG, I see the h2, h3, h4 headings in the WYSIWYG. When I save, they each get downgraded, so it starts at h3. Is this intentional? There is no h1 which makes sense because of the page title, but maybe multiple h2 headings is fine.

All other formatting seems ok.

When I try to apply the "Remove formatting" (highlight the text, click icon) nothing seems to happen.

When I follow the test instructions and create a new job or training, and enter stuff directly in the WYSIWYG, save and re-edit without changing anything, I can confirm there is no new entry in the revisions.
When I do the same but paste in HTML, save and re-edit without changing, it seems to create a new revision but I can't see anything having changed. This seems to only happen the first time, not sure if it's relevant.

I think injecting the styles is a good idea.

orakili:
@left23 I've fixed the heading hierarchy issue. There was an issue in the node templates with the maximum heading level for the body fields (leftover of D7 I think where the page title was a h2 not h1).

I also added a CKEditor plugin to convert the headings when copying/pasting (ex: h1 -> h2 etc.)

There is also a fix for RW-273 for the ckeditor link dialog submit button. Looks like it's not a RW issue: https://www.drupal.org/project/drupal/issues/3089751. The fix in https://github.com/UN-OCHA/rwint9-site/pull/198/commits/c65c5ca7903289c869cfd7bd5f030039a07711d7 could be moved to the base theme, what do you think?

Finally, I injected the subtheme `styles.css` into the editor and it seems to work fine.

left23 (changes requested):
@orakili I tested initially in Firefox 95 (linux) and I am unable to see the textarea. It appears to be a JS error.


In Chrome it's fine so I did the rest of the testing there.

I can confirm the markup gets adjusted to the appropriate headings: Starting at h2 for Job Description and Body textareas, and h3 for How to Apply and How to Register. 

The rest of the formatting works as expected. I can confirm the typography styles are working and the link modal is fixed. 

I will create an item in the CD backlog. I think it's not currently necessary to add this to the base theme as we've verified that other sites don't have this problem due to minimal or no manipulation of node forms. It's good to keep a record of it as a start.

orakili:
@left23 Thanks for finding this issue. I only tested in webkit/blink based browsers and ckeditor enables the pasteFilter by default in those browsers but not in others that's why there was the error in firefox for example.

I've committed a fix and that should work cross browsers now, hopefully.

left23 (approved):

PR #199, opened 2021-12-10
RW-25  File widget styling

left23:
1. Small style and layout improvements to the File Widget
2. Adds a custom toggle arrow icon to the Actions `<details>` element
3. Includes RW-271 error message consistency (relies on [base theme branch updates](https://github.com/UN-OCHA/common_design/pull/254)) `composer update unocha/common_design`
4. Removes datetime-wrapper template override as this is now in the base theme
5. Improve the UI for the Revisions actions (Revert, Delete) in the revisions table (in base theme) and add padding to the revision confirmation pages


html/themes/custom/common_design_subtheme/common_design_subtheme.libraries.yml
```diff
@@ -132,6 +132,8 @@ rw-form:
     theme:
       components/rw-form/rw-form.css: {}
       components/rw-form/rw-form--widget.css: {}
+  dependencies:
+    - common_design/cd-form
```
left23, line 136: I've added this as a dependency because I noticed it did not load on the revision confirmation pages, despite `rw-form` component loading. I want to run VRT to confirm there are no negative affects.

left23:
@orakili You can review this whenever. Note the base branch, so maybe it makes sense to review it as part of the larger RW-25

orakili:
@left23 I've merged that in the feature/RW-25 branch and deployed it on dev.

It looks good. 

I'm wondering if we can make the row more compact (with a smaller thumbnail) like on the D7 site:

Screen Shot 2021-12-15 at 23 13 06

and maybe also "fix" the "actions" column to reduce layout shifts.

What do you think?

left23:
I've copied what styles were relevant from D7. I also renamed the select option "No preview" to "none" for reduce the necessary width. It matches the language in the "Preview page rotation" options too. @orakili feel free to revert if you dislike.

Selection_190

I don't think fixing the width of the Actions column makes sense. It needs a lot of space, and it would make it more obvious, and also reduce the available space for the other fields which will be used more frequently. Having it narrow and shifting everything when it is open also lends itself to making it clear it as a "destructive" action. Unless we hear strong opinions to the contrary, I think we should leave it.

orakili (approved):
@left23 Thanks.

 I managed to get the tabledrag working (RW-269) and I've added the field label ("Attachments"), renamed the rotation options and removed one level of borders to make it more compact:

Screen Shot 2021-12-16 at 22 18 27

Screen Shot 2021-12-16 at 22 21 31

I think that's close enough to the D7 design, so I'm merging that in the RW-25 branch.

PR #202, opened 2021-12-16
[RW-283] Drupal code and modules update

orakili:
Ticket: RW-283

This updates to Drupal 9.3 which contains the work on the bundle classes so we can remove the huge patch we were using.

This PR also updates the modules.

There is a ton of file changes from https://github.com/UN-OCHA/rwint9-site/pull/202/commits/6d1e37f0a8752613e75a45988ddd2abab5a659f3 but it's only some property re-ordering in the config files.


orakili:
Tested all the pages I could think off and didn't encounter any issues so merging this and will update other PRs otherwise it's going to be difficult to switch from one to the other.

PR #207, opened 2021-12-29
add process and destination configuration for migrations

lazysoundsystem:
Not sure what effect these errors have on the migration itself, but I was unable to visit pages like https://dev.reliefweb-int.ahconu.org/admin/structure/migrate/manage/reliefweb_node/migrations/reliefweb_node__book/process and https://dev.reliefweb-int.ahconu.org/admin/structure/migrate/manage/reliefweb_node/migrations/reliefweb_node__book/destination as the configuration is missing.

```
The website encountered an unexpected error. Please try again later.
TypeError: Drupal\migrate\Plugin\Migration::getProcessNormalized(): Argument #1 ($process) must be of type array, null given, called in /srv/www/html/core/modules/migrate/src/Plugin/Migration.php on line 630 in Drupal\migrate\Plugin\Migration->getProcessNormalized() (line 375 of core/modules/migrate/src/Plugin/Migration.php). 
```

This PR fixes those.

### DIFF
--- a/config/migrate_plus.migration.reliefweb_media__image_topic.yml
+++ b/config/migrate_plus.migration.reliefweb_media__image_topic.yml
@@ -18,7 +18,11 @@ source:
   field: field_term_image
   entity_type: node
   bundle: topics
-process: null
+process:
+  field_media_image/target_id:
+    plugin: skip_on_empty
+    method: process
+    source: fid
 destination:
   default_bundle: image_topic
 migration_dependencies:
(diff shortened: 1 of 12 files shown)


orakili:
@lazysoundsystem Thanks but those changes are not needed because the field mapping etc. is in the shared configuration of the groups those migrations are based on and the config files are replaced regularly so those changes will be lost.

Also, all the migration stuff is done via the drush commands not the interface, so closing this.

PR #210, opened 2022-01-03
[RW-295] fix small typos

lazysoundsystem:
https://humanitarian.atlassian.net/browse/RW-295

All are in code comments apart from the 'valyes' variable.


orakili (approved):
Thanks!

PR #220, opened 2022-01-18
RW-314 RW-315 Style changes based on recent testing

left23:
RW-314 add rule to override external link icon in pseudo element on disaster map icon

RW-315 replace SCSS colour variable with RGB CSS custom property for colour, so active menu RGBA value applies when mobile menu items are active.


left23:
@lazysoundsystem Adding you as a reviewer as well for awareness.

PR #221, opened 2022-01-19
RW-316 Pager

left23:
This moves the pager outside the list section element and adjusts the styles for its display. It also changes the pager nav title from a h3 to h2 to reflect the change in hierarchy.


orakili (approved):
Thanks!

PR #231, opened 2022-01-27
Frontend adjustments based on testing

left23:
RW-285 (See issue - this was undone after some discussion because a better solution was found)
RW-323
RW-324
RW-321
RW-322


orakili (approved):
Thanks.

PR #233, opened 2022-01-28
[OPS-8024] Allow lua access to an extra env var without a custom config file.

cafuego:
Since there is an extra build step to install the lua uuid module anyway, its config may as well live in the same Dockerfile.


orakili (approved):
Thanks.

PR #238, opened 2022-02-02
RW-343 add clearfix rule to clear preceding floated element

left23:
RW-343 This looks after the display with the WYSIWYG only, and keeps the rule from RW-322 which does the same thing for the autocomplete widget.


orakili (approved):
Thanks, merging as it's fixing the wysiwyg issue but I think adding classes to the `label` and `legend` elementsto which we are adding the guideline buttons  (see [here](https://github.com/UN-OCHA/rwint9-site/blob/f3f87e3f583996d10e15593508fa79b72bbe4ba3/html/themes/custom/common_design_subtheme/components/rw-guidelines/guidelines-json.js#L176-L181)) would be make the overall styling easier to maintain. What do you think?

PR #261, opened 2022-03-08
RW-177 Preview refactor

left23:
This is dependent on a base theme release that includes this PR https://github.com/UN-OCHA/common_design/pull/286

This has now been merged to `main` so if this https://github.com/UN-OCHA/rwint9-site/pull/264 is merged to develop, then this PR would work as expected.

I've removed all CSS rules and template overrides that are now in the base theme, and set the `rw-preview` library to extend the base theme library so we can retain the changes in colour and hiding the view mode select specifically for RW.

Note: I came across a bug which I logged here https://humanitarian.atlassian.net/browse/CD-370


orakili (approved):
Thanks!

PR #263, opened 2022-03-09
RW-406 Guideline menu item with icon

left23:
Expected result:

Selection_016

Selection_017


orakili (approved):
@left23 working well. Thanks!

For info, I created a CD ticket and PR to include the sprite via a template rather than the html preprocess hook: https://humanitarian.atlassian.net/browse/CD-373

Also created https://humanitarian.atlassian.net/browse/RW-415 as a low priority follow-up ticket to remove unused icons from the sprite.

PR #264, opened 2022-03-09
RW-405 update theme

left23:
- Add nvm file to sub theme with correct node version
- Set base theme tag to v6


orakili (approved):
Thanks!

PR #279, opened 2022-04-12
RW-438 update base theme and SVG sprite icon partial

left23:
- Upgrade base theme to v7 https://github.com/UN-OCHA/common_design/releases/tag/v7.0.0
- Removes the SVG icon sprite from being attached in the sub theme preprocess, and instead uses a template override
There should be visual changes, and the social media icons should continue to display in the footer.


orakili (approved):
Seems to be working well. Thanks.

I've replaced the html.html.twig template with the one from the `common_design/common_design_subtheme` one which just contains the block definition to  add the subtheme icon template.

PR #284, opened 2022-04-20
[RW-467] Only add optional mark before rendering form elements

orakili:
Ticket: RW-467

This PR delays the addition of the "optional" mark so that other places using the form element title (ex: to set an error message) see the original one.

Due to the way drupal builds the form elements we need to pass around the optional information before we can finally alter the title:

1. Process function to add an `after build` callback to the relevant form elements
2. After build callback to add the `pre render` callbacks to add the optional mark
3. Pre render callback to add the "optional" mark

## Testing

Create a training, for example, and leave the "registration deadline" empty. Without this PR, the error message at the top should show "Registration deadline - optional". With the PR, the message should show "Registration deadline" only.

Other optional fields like "Theme" should still display the "- optional" text next to the title.

### DIFF
--- a/html/themes/custom/common_design_subtheme/common_design_subtheme.theme
+++ b/html/themes/custom/common_design_subtheme/common_design_subtheme.theme
@@ -5,13 +5,12 @@
  * Preprocess functions for the Common Design subtheme.
  */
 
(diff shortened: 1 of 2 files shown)


left23 (approved):
Works, thanks!

PR #337, opened 2022-05-18
[OPS-8307] Add the memcache services, so they can be used. Wheee!

cafuego:
This depends on https://github.com/UN-OCHA/tools-ansible/pull/851 (which is applied)


orakili (approved):
@cafuego Do we need to update `config/core.extensions.yml` to enable the memcache module or does ansible do that for us?

cafuego:
Yes, but the updated settings.php needs to be present on disk first, I think. I'm not sure what the module will do when enabled and not configured.

PR #339, opened 2022-05-19
[RW-554] Disable nginx rules related to the docstore

orakili:
Disable nginx rules related to the docstore as it is not used to avoid upstream not found issues when there is no connected docstore instance.


cafuego (approved):
Looks and sounds reasonable.

PR #353, opened 2022-06-01
RW-573 Checkbox layout on unsubscribe form for logged out users

left23:
- add selector for unsubscribe form checkbox items
- space out cancel button from save button

As I cannot see this page locally, I did what I could in the browser using Prod.
Expected results:
Selection_012


orakili (approved):
Thanks!

PR #354, opened 2022-06-02
refactor: Do not use state State api, as it cannot be populated via settings.php

cafuego:
Refs: OPS-8349


orakili (approved):
Tested locally and the settings are picked up properly.

PR #363, opened 2022-06-07
RW-573 Unsubscribe form styles in global library

left23:
I moved the unsubscribe form rules from `rw-user` component and added them to `rw-form` which is part of the global library. The first attempt #353 didn't work because the library isn't loaded onto that page and since I cannot view this page locally, using the global library seems like the most straight-forward.


lazysoundsystem (approved):
Discussed at standup - including this change now to join today's deployment.

PR #364, opened 2022-06-07
bug: Unsubscribe countries

attiks:
Refs: #RW-582


left23 (commented):
@attiks I checked out this branch and cleared cache but I don't get the intended results.
When remove a country and save the subscriptions page, the country isn't removed.
Are there additional steps I need to do to test this?

attiks:
Just fixed a bug when removing the last country

left23 (approved):
Thanks, works now.

PR #371, opened 2022-06-14
[WIP][OPS-8360] feat: add script to generate lighter DB dump

orakili:
Refs: OPS-8360

Script that generates a DB sql dump file, only keeping the most recent jobs, training and reports.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,7 +31,7 @@ RUN cp -a docker/settings.php docker/services.yml docker/memcache.services.yml h
 ################################################################################
 
 # Generate the image.
-FROM public.ecr.aws/unocha/php-k8s:8.0.19-r1-NR-202206-01
+FROM public.ecr.aws/unocha/php-k8s:8.0-NR-stable
 
 ARG VCS_REF
 ARG VCS_URL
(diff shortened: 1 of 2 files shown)


orakili:
@cafuego This PR contains a script that generates a DB dump of ~750MB (not zipped) for a DB of ~2.5GB after import.

It does so by exporting only 1000 reports, 1000 jobs and 1000 training (most recent). All the other node types, terms, guidelines etc. are preserved as they are necessary for the site well behavior.

It's meant to be adapted to use in a jenkins job. Please let me know what you think.

For info, it takes a couple of minutes to export and ~8 minutes to import on my local.

docker/Dockerfile
```diff
@@ -54,7 +54,8 @@ LABEL info.humanitarianresponse.build.date=$BUILD_DATE \
       info.humanitarianresponse.build.vcs-url=$VCS_URL \
       info.humanitarianresponse.build.vcs-ref=$VCS_REF
 
-RUN mkdir -p /etc/nginx/custom && \
+RUN apk add -U icu-data-full && \
+    mkdir -p /etc/nginx/custom && \
```
cafuego, line 58:
```suggestion
RUN mkdir -p /etc/nginx/custom && \
```

cafuego:
Sounds good. I'll see if I can integrate it into a custom Jenkins job, so Ansible doesn't wipe it (and I won't need to add it to Ansible as a snowflake). I think it also means I can make a change to have Ansible not generate backup jobs for dev and demo, I'm not sure how valuable it would be to have backup dumps of reduced size databases when there is a reduced size prod one available anyway.

orakili:
Merging to keep that as reference.

PR #376, opened 2022-06-20
feat: Most read

attiks:
Refs: #RW-23

### DIFF
--- a/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
+++ b/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
@@ -123,6 +123,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension20' => (string) $entity->getModerationStatusLabel(),
         'dimension21' => $entity->id(),
         'dimension22' => $entity->uuid(),
+        'dimension23' => reliefweb_analytics_get_terms_id($entity, 'field_country', $cache_metadata),
+        'dimension24' => reliefweb_analytics_get_terms_id($entity, 'field_disaster', $cache_metadata),
       ];
       break;
 
@@ -297,6 +299,30 @@ function reliefweb_analytics_get_terms(FieldableEntityInterface $entity, $field,
   if ($entity->hasField($field) && !$entity->get($field)->isEmpty()) {
     foreach ($entity->get($field)->referencedEntities() as $term) {
       $terms[] = (string) $term->getUntranslated()->label();
+      $cache_metadata->addCacheTags($term->getCacheTags());
+    }
+  }
+  return implode(', ', $terms);
+}
+
+/**
+ * Get the term labels from an entity reference field.
+ *
+ * @param \Drupal\Core\Entity\FieldableEntityInterface $entity
+ *   Entity.
+ * @param string $field
+ *   Field.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cache_metadata
+ *   The accumulated cache metadata.
+ *
+ * @return string
+ *   Comma separated list of term names.
+ */
+function reliefweb_analytics_get_terms_id(FieldableEntityInterface $entity, $field, CacheableMetadata $cache_metadata) {
+  $terms = [];
+  if ($entity->hasField($field) && !$entity->get($field)->isEmpty()) {
+    foreach ($entity->get($field)->referencedEntities() as $term) {
+      $terms[] = (string) $term->id();
       $cache_metadata->addCacheTags($term->getCacheTags());
     }
   }
--- a/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
+++ b/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
@@ -121,6 +121,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         /*'dimension11' => reliefweb_analytics_get_terms($entity, 'field_vulnerable_groups'),*/
         'dimension13' => reliefweb_analytics_format_date($entity, 'created'),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -136,6 +138,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension16' => reliefweb_analytics_get_terms($entity, 'field_job_experience', $cache_metadata),
         'dimension17' => reliefweb_analytics_get_terms($entity, 'field_career_categories', $cache_metadata),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -152,6 +156,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension18' => reliefweb_analytics_get_terms($entity, 'field_training_type', $cache_metadata),
         'dimension19' => reliefweb_analytics_get_terms($entity, 'field_training_format', $cache_metadata),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -159,6 +165,8 @@ function reliefweb_analytics_get_entity_dimensions() {
       $dimensions = [
         'dimension1' => 'Topic',
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -166,6 +174,8 @@ function reliefweb_analytics_get_entity_dimensions() {
       $dimensions = [
         'dimension1' => 'Blog post',
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -173,6 +183,8 @@ function reliefweb_analytics_get_entity_dimensions() {
       $dimensions = [
         'dimension1' => 'Book page',
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -182,6 +194,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension1' => 'Country',
         'dimension6' => (string) $entity->label(),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -194,6 +208,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension10' => reliefweb_analytics_get_terms($entity, 'field_disaster_type', $cache_metadata),
         'dimension13' => reliefweb_analytics_format_date($entity, 'field_disaster_date'),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
 
@@ -203,6 +219,8 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension6' => reliefweb_analytics_get_terms($entity, 'field_country', $cache_metadata),
         'dimension7' => (string) $entity->label(),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
       ];
       break;
   }
--- a/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
+++ b/html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
@@ -121,9 +121,6 @@ function reliefweb_analytics_get_entity_dimensions() {
         /*'dimension11' => reliefweb_analytics_get_terms($entity, 'field_vulnerable_groups'),*/
         'dimension13' => reliefweb_analytics_format_date($entity, 'created'),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
-        'dimension21' => $entity->id(),
-        'dimension22' => reliefweb_analytics_get_terms_id($entity, 'field_primary_country', $cache_metadata),
-        'dimension23' => reliefweb_analytics_get_terms_id($entity, 'field_disaster', $cache_metadata),
       ];
       break;
 
@@ -139,7 +136,6 @@ function reliefweb_analytics_get_entity_dimensions() {
         'dimension16' => reliefweb_analytics_get_terms($entity, 'field_job_experience', $cache_metadata),
         'dimension17' => reliefweb_analytics_get_terms($entity, 'field_career_categories', $cache_metadata),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
-        'dimension21' => $entity->id(),
       ];
       break;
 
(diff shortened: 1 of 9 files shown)


html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+   * Most read for disasters.
+   *
+   * @command reliefweb_analytics:disasters
+   * @usage reliefweb_analytics:disasters
+   *   Send emails.
+   * @validate-module-enabled reliefweb_analytics
+   * @aliases reliefweb-mostread-disasters
+   */
+  public function disasters() {
+    // Load terms having a job URL.
```
orakili, line 132:
```suggestion
```

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+    ]));
+
+    foreach ($response->getRows() as $row) {
+      if (isset($row->getDimensionValues()[1]) && !empty($row->getDimensionValues()[1]->getValue()) && $row->getDimensionValues()[1]->getValue() !== '(not set)') {
+        if (!isset($results[$row->getDimensionValues()[1]->getValue()])) {
+          $results[$row->getDimensionValues()[1]->getValue()] = [];
+        }
+        if (count($results[$row->getDimensionValues()[1]->getValue()]) < 5) {
+          $results[$row->getDimensionValues()[1]->getValue()][] = $row->getDimensionValues()[0]->getValue();
+        }
```
orakili, line 334: I'm not sure this will produce the expected result because some reports can be tagged with multiple disasters in that case `$row->getDimensionValues()[1]` may be a comma separated list of term labels (if using `customEvent:content_report_disaster`) or of term Ids (if using `dimension24` once available).

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+        'string_filter' => new StringFilter([
+          'value' => '.',
+          'match_type' => MatchType::PARTIAL_REGEXP,
+        ]),
+      ]),
+    ]);
+    $payload = $this->buildPayload($filter);
+
+    // Add dimension for disaster.
+    $payload['dimensions'][] = new Dimension(['name' => 'customEvent:content_report_disaster']);
```
orakili, line 460: I'm guessing this is to be replaced with `dimension24`.

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+   * @aliases reliefweb-mostread-disasters-all
+   */
+  public function disastersAll() {
+    $results = [];
+    $parameters = $this->getAllDisastersPayload();
+    $combined = $this->fetchGa4DataCombined($parameters);
+
+    if (!empty($combined)) {
+      foreach ($combined as $disaster => $data) {
+        $results[$disaster] = [
```
orakili, line 170: Isn't `disaster` here the disaster label as `getAllDisastersPayload` uses the `content_report_disaster` dimension? If that's the case, I'm guessing there should be some lookup to do to retrieve the ID?

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+      $this->updateCsv($results);
+    }
+  }
+
+  /**
+   * Most read for countries.
+   *
+   * @command reliefweb_analytics:countries
+   * @usage reliefweb_analytics:countries
+   *   Send emails.
```
orakili, line 91:
```suggestion
   *   Retrieve the most read reports for each country.
```

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+      $this->updateCsv($results);
+    }
+  }
+
+  /**
+   * Most read for disasters.
+   *
+   * @command reliefweb_analytics:disasters
+   * @usage reliefweb_analytics:disasters
+   *   Send emails.
```
orakili, line 127:
```suggestion
   *   Retrieve the most read reports for each disaster.
```

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+      $this->updateCsv($results);
+    }
+  }
+
+  /**
+   * Most read for disasters.
+   *
+   * @command reliefweb_analytics:disasters-all
+   * @usage reliefweb_analytics:disasters-all
+   *   Send emails.
```
orakili, line 159:
```suggestion
   *   Retrieve the most read reports for each disaster.
```

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+      }
+    }
+
+    if (!empty($results)) {
+      $this->updateCsv($results);
+    }
+  }
+
+  /**
+   * Most read for disasters.
```
orakili, line 155:
```suggestion
   * Most read for disasters using a single request.
```

orakili (changes requested):
That looks promising, thanks!

How are the results with the 100,000 reports query? Do they cover most disasters?

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+    $this->entityTypeManager = $entity_type_manager;
+    $this->loggerFactory = $logger_factory;
+  }
+
+  /**
+   * Most read for homepage.
+   *
+   * @command reliefweb_analytics:homepage
+   * @usage reliefweb_analytics:homepage
+   *   Send emails.
```
orakili, line 63:
```suggestion
   *   Retrieve the most read reports for the homepage.
```

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+    }
+  }
+
+  /**
+   * Get GA4 clinet.
+   */
+  protected function getGa4Client() {
+    static $client = NULL;
+    if (!$client) {
+      $client = new BetaAnalyticsDataClient(['credentials' => '/var/www/credentials.json']);
```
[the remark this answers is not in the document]
attiks: I'll make it config, so ansible can decide

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,468 @@
...
+   *   Send emails.
+   * @validate-module-enabled reliefweb_analytics
+   * @aliases reliefweb-mostread-disasters
+   */
+  public function disasters() {
+    // Load terms having a job URL.
+    $query = $this->entityTypeManager->getStorage('taxonomy_term')->getQuery();
+    $query->condition('vid', 'disaster');
+    $tids = $query->execute();
+    $disasters = $this->entityTypeManager->getStorage('taxonomy_term')->loadMultiple($tids);
```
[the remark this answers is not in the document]
attiks: I'll update it

attiks:
> How are the results with the 100,000 reports query? Do they cover most disasters?

Yes, total of results is 210.000, but number 100.000 has only 2 hits, so should be fine

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
attiks: The dimensions from `reliefweb_analytics_page_attachments` should be expanded again so not `dimension9 = 1,2,3` but `dimension9=1`, `dimension9=2`, `dimension9=3` which is supported by GA

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
attiks: Yes, left it for testing

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
attiks: I stand corrected, dimensions are either number or string, so I'll expand in code

orakili (commented):
A couple of things that need to be added:

**Term name -> ID conversion**

Should we add code to do the term Name -> ID lookup in this drush command (and remove it when starting using the new dimensions)? 

With less than 4000 terms, I guess the mapping can be loaded in one DB query so that should be relatively performant.

Otherwise that could be done in https://github.com/UN-OCHA/rwint9-site/blob/a1e9785b41b991ef91f7832e73c12a039a0ad6cb/html/modules/custom/reliefweb_entities/src/SectionedContentTrait.php#L276

(I would propose to do that in the drush command.)

**Code for homepage**

This: https://github.com/UN-OCHA/rwint9-site/blob/a1e9785b41b991ef91f7832e73c12a039a0ad6cb/html/modules/custom/reliefweb_homepage/src/Controller/Homepage.php#L272 needs to be updated to do something similar to https://github.com/UN-OCHA/rwint9-site/blob/a1e9785b41b991ef91f7832e73c12a039a0ad6cb/html/modules/custom/reliefweb_entities/src/SectionedContentTrait.php#L276.

html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
```diff
@@ -121,6 +121,10 @@ function reliefweb_analytics_get_entity_dimensions() {
         /*'dimension11' => reliefweb_analytics_get_terms($entity, 'field_vulnerable_groups'),*/
         'dimension13' => reliefweb_analytics_format_date($entity, 'created'),
         'dimension20' => (string) $entity->getModerationStatusLabel(),
+        'dimension21' => $entity->id(),
+        'dimension22' => $entity->uuid(),
```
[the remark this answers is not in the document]
attiks: No

attiks:
> Term name -> ID conversion

We can use `reliefweb_analytics:countries` and `reliefweb_analytics:disasters` so we don't have to do the lookup

We do need path alias mapping to entity_id

html/modules/custom/reliefweb_analytics/reliefweb_analytics.module
```diff
@@ -121,6 +121,10 @@ function reliefweb_analytics_get_entity_dimensions() {
...
+        'dimension23' => reliefweb_analytics_get_terms_id($entity, 'field_country', $cache_metadata),
```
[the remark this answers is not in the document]
attiks: Updated

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,594 @@
...
+            'value' => 'Report',
+          ]),
+        ]),
+      ]),
+      new FilterExpression([
+        'filter' => new Filter([
+          'field_name' => 'pagePath',
+          'string_filter' => new StringFilter([
+            'match_type' => MatchType::BEGINS_WITH,
+            'value' => '/report',
```
orakili, line 459: @attiks Sorry, I just noticed that map reports have their URL path prefix with `/map` not `/report` (ex: https://reliefweb.int/map/mongolia/mongolia-drought-map-ten-day-period-starting-11-june-2022-enmn) so the filter should check both. Hopefully that will not increase much the number of tokens that the query uses.

attiks: Or we remove it, there's already a filter on content type?

orakili: The `/report` filter was there initially because there were some `/node/1234566/edit` (from editors) URLs "polluting" the results but it's probably rare enough to not care.

So, yes, let's remove it and see how it goes. That may reduce the number of tokens required.

attiks: Removed

html/modules/custom/reliefweb_homepage/src/Controller/Homepage.php
```diff
@@ -270,18 +270,49 @@ public function getHeadlinesApiPayload($limit = 8) {
...
+    }
+
+    // Close the file.
+    if (is_resource($handle)) {
+      @fclose($handle);
+    }
+
+    // Generate the query with the most read report ids.
+    if (empty($ids)) {
+      return [];
```
orakili, line 296: Sorry, it was not indicated but to have the same behavior as RW7:

- If there is most read data, the title should be `$this->t('Most Read <span>(last 24 hours)</span>')`.
- If there is no data from the most read CSV file, this should use the payload to get the latest updates (code before this change) and use `$this->t('Latest Updates')` as title for the block.

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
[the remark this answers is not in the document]
orakili: Bumping.

html/modules/custom/reliefweb_analytics/src/Command/ReliefwebMostReadCommand.php
```diff
@@ -0,0 +1,599 @@
...
+    $payload = $this->buildPayload();
+
+    $payload['dateRanges'] = [
+      new DateRange([
+        'start_date' => $this->state->get('reliefweb_analytics_homepage_start_date', 'yesterday'),
+        'end_date' => 'today',
+      ]),
+    ];
+
+    return $payload;
```
orakili, line 514: Note that for the homage only **2** reports are needed.

html/modules/custom/reliefweb_homepage/src/Controller/Homepage.php
```diff
@@ -270,18 +270,49 @@ public function getHeadlinesApiPayload($limit = 8) {
...
+    }
+
+    // We reverse the ids to add the boost (higher boost = higher view count).
+    foreach (array_reverse($ids) as $index => $id) {
+      $ids[] = $id . '^' . ($index * 10);
```
orakili, line 301:
```suggestion
      $ids[$index] = $id . '^' . ($index * 10);
```

We need to override the value with the boosted one.

html/modules/custom/reliefweb_homepage/src/Controller/Homepage.php
```diff
@@ -270,18 +270,49 @@ public function getHeadlinesApiPayload($limit = 8) {
...
+    // heavy.
+    $handle = @fopen('public://most-read/most-read.csv', 'r');
+    if ($handle === FALSE) {
+      return [];
+    }
+
+    // Find the line corresponding to the entity id.
+    while (($row = fgetcsv($handle, 100)) !== FALSE) {
+      if (count($row) === 2 && $row[0] == 'front') {
+        $ids = explode(',', $row[1]);
```
orakili, line 284:
```suggestion
        $ids = array_slice(explode(',', $row[1]), 0, $limit);
```

orakili (commented):
Looks good, thanks. I've added some comments for the homepage.

orakili (approved):
I think we are good to go. Thanks a lot for the work on that feature.

We may need to adjust the results after we receive feedback (for example to boost recent content etc.) but that will be for another ticket/PR if the need arises.

PR #378, opened 2022-06-20
feat: Fetch data

attiks:
Refs: #RW-23


attiks:
See https://github.com/UN-OCHA/rwint9-site/pull/376

PR #301, opened 2022-05-05
RW-482 Remove SASS from sub theme

left23:
Merge after https://github.com/UN-OCHA/rwint9-site/pull/345 because this deletes files.

1. Remove theme-related docker build step
2. Replace styles.css with brand stylesheet for global styles
3. Transfer customized sass variables to CSS custom properties in brand.css
4. Define new library for CD-related CSS `rw-cd-header-footer`
5. Replace sass partials with css files
6. Convert sass to CSS, and replace vars with custom property syntax
7. Delete sass partials

PR #379, opened 2022-06-20
RW-589 Soft Footer with Mailchimp

left23:
This branches off #301 

Selection_072


orakili (approved):
Thanks, it looks good. 

I just added translations of the different textual elements.

PR #382, opened 2022-06-21
feat: Intermediate page for login

attiks:
Refs: #RW-599


orakili:
Cherry picked so we can merge to develop without the most read work: https://github.com/UN-OCHA/rwint9-site/pull/385

PR #388, opened 2022-06-27
feat: No more redis, also use memcache in tests

cafuego:
Refs: OPS-8319


cafuego:
The test email fix (as listed in `drush help si`) doers not actually fix the mail error on install, but on the other hand it also does not break the tests, so I'm leaving it in.

orakili (approved):

PR #389, opened 2022-06-27
RW 601: Widget with description

attiks:


composer.json
```diff
@@ -32,6 +32,7 @@
         "drupal/core-composer-scaffold": "^9.3",
         "drupal/core-dev": "^9.3",
         "drupal/core-recommended": "^9.3",
+        "drupal/form_options_attributes": "^1.2",
```
orakili, line 35: RW has already an implementation to handle adding option attributes: https://github.com/UN-OCHA/rwint9-site/blob/f563644a643e80963d517beda2094fd58743a57d/html/modules/custom/reliefweb_form/reliefweb_form.module#L29-L30 using the `#option_attributes` property, that do something similar.

config/core.entity_form_display.node.training.default.yml
```diff
@@ -185,7 +185,7 @@ content:
         'source:field_user_posting_rights': 0
     third_party_settings: {  }
   field_theme:
-    type: options_buttons
+    type: reliefweb_options
```
orakili, line 188: I don't see the code for the widget in the PR but if that's not already the case, please have a look at the https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebEntityReferenceSelect.php widget which adds widget options to add fields as data attributes to select option elements.

This widget is for `select` but could be adapted easily enough for `checkboxes/radios`.

It's used for example for the disaster field on the report form to add the disaster status as a data attribute which is then used to add a colored border to the suggestions.

composer.json
attiks: Module removed

config/core.entity_form_display.node.training.default.yml
attiks: Missing file added

attiks:
PS: Labels in popup can be clicked to control radfio/checkbox

config/core.extension.yml
```diff
@@ -15,6 +15,7 @@ module:
   ctools: 0
   datetime: 0
   datetime_range: 0
+  devel: 0
```
orakili, line 18: Devel module is enabled via config_split: https://github.com/UN-OCHA/rwint9-site/blob/c41fd92c2e6ecb511bbb0001b616099d63a232ac/config/config_split.config_split.config_dev.yml#L14

composer.lock
```diff
@@ -15248,5 +15298,5 @@
         "php": ">=8.0"
     },
     "platform-dev": [],
-    "plugin-api-version": "2.3.0"
+    "plugin-api-version": "2.2.0"
```
orakili, line 15301: Doesn't matter much but the version of composer used for the current site images in 2.3.x.

html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebOptions.php
```diff
@@ -0,0 +1,119 @@
...
+ * @FieldWidget(
+ *   id = "reliefweb_options",
+ *   module = "reliefweb_fields",
+ *   label = @Translation("ReliefWeb Options with description"),
+ *   multiple_values = true,
+ *   field_types = {
+ *     "list_integer",
+ *     "list_string",
+ *     "list_float",
+ *     "boolean",
```
orakili, line 23: The code seems to only work for entity reference field types so I think that should be removed.

html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebOptions.php
```diff
@@ -0,0 +1,119 @@
...
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $element = OptionsWidgetBase::formElement($items, $delta, $element, $form, $form_state);
+
+    $options = $this->getOptions($items->getEntity());
+    $selected = $this->getSelectedOptions($items);
+
+    // Load all terms, add description.
+    $entity_type_id = $this->getReferencedEntityTypeId();
```
orakili, line 42: The code below seems to only work for the taxonomy terms so maybe it can bail out earlier with something like:

```php
if ($entity_type_id !== 'taxonomy_term') {
   return parent::formElement($items, $delta, $element, $form, $form_state);
}
```

html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebOptions.php
```diff
@@ -0,0 +1,119 @@
...
+    $entities = [];
+    foreach ($bundles as $bundle) {
+      if ($entity_type_id === 'taxonomy_term') {
+        $entities += $entity_type_manager->getStorage($entity_type_id)->loadByProperties([
+          'vid' => $bundle,
+        ]);
+      }
+      else {
+        $entities += $entity_type_manager->getStorage($entity_type_id)->loadByProperties([
+          'bundle' => $bundle,
```
orakili, line 55: Media or nodes uses `type` not `bundle` for example. Also, as there is only one `entity_type_id`, maybe the whole foreach block can be simplified:


```php
$entities = $entity_type_manager->getStorage($entity_type_id)->loadByProperties([
  $this->getEntityTypeBundleField($entity_type_id) => $bundles,
]);
```

orakili (commented):
Thanks, that seems to be working well.

The `css` and possibly the `js` file will probably need to be overridden in the subtheme to add the necessary classes so that the styling from the `rw-guidelines` component can be re-used.

Also the fact that clicking on the terms in the popup select the guidelines is quite confusing in terms of UX and accessibility as there is no feedback when doing so and the selection only appears after closing the popup. I think we should avoid that and just let the popup be help information without interactivity.

html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebOptions.php
attiks: True

attiks:
> Also the fact that clicking on the terms in the popup select the terms is quite confusing in terms of UX and accessibility as there is no feedback when doing so and the selection only appears after closing the popup. I think we should avoid that and just let the popup be help information without interactivity.

I'm looking how I can add a visual indicator

config/core.entity_form_display.node.training.default.yml
```diff
@@ -197,7 +197,7 @@ content:
     settings: {  }
     third_party_settings: {  }
   field_training_format:
-    type: options_buttons
+    type: reliefweb_options
```
orakili, line 200:
```suggestion
    type: options_buttons
```

Not needed for the training format. It's only 2 choices:  online or on site and the description for those terms is basically the term label. It's also not required in the RW-601 ticket.

html/modules/custom/reliefweb_fields/components/reliefweb-options/reliefweb-options.js
```diff
@@ -0,0 +1,228 @@
...
+          if (source.checked) {
+            checkbox.setAttribute('checked', 'checked');
+          }
+          checkbox.setAttribute('name', 'shadow-' + source.getAttribute('name'));
+          checkbox.setAttribute('id', 'shadow-' + term.for);
+          checkbox.setAttribute('data-for', term.for);
+          checkbox.addEventListener('change', function (e) {
+            let real = document.getElementById(this.getAttribute('data-for'));
+            real.checked = this.checked;
+          });
```
orakili, line 143: The problem with that approach is that it doesn't respect the maximum of allowed terms. For example by selecting the professional function terms in the popup, I could exceed the limit:

Screen Shot 2022-06-30 at 15 50 06

That's also going to cause issues with the `career categories` for jobs where the `theme` options become disabled for some career categories (for example themes are not allowed when selecting "Administration/Finance"):

Screen Shot 2022-06-30 at 15 53 21

I haven't tested but maybe another way to handle that could be to transform the original field wrapper into the popup. By that I mean, adding a class to the fieldset wrapper that would result in it being styled as a popup and showing the description via the a `::after` for example (a bit like the option 2 in the RW-601 ticket but with the field styled as a popup). If that works, that could reduce the complexity of the javascript as well.

html/modules/custom/reliefweb_fields/src/Plugin/Field/FieldWidget/ReliefWebOptions.php
```diff
@@ -0,0 +1,109 @@
...
+        // Radio buttons need a scalar value. Take the first default value, or
+        // default to NULL so that the form element is properly recognized as
+        // not having a default value.
+        '#default_value' => $selected ? reset($selected) : NULL,
+        '#options' => $options,
+        '#option_attributes' => $option_attributes,
+      ];
+    }
+
+    $element['#attached']['library'][] = 'reliefweb_fields/reliefweb-options';
```
orakili, line 83: I wonder if we can add a class to the field wrapper here to ease the selection in the javascript. Something like `$element['#wrapper_attributes']['class'] = 'data-with-term-descriptions';` (not tested).

attiks: `$element['#attributes']['class'][] = 'data-with-term-descriptions';` works

html/modules/custom/reliefweb_fields/components/reliefweb-options/reliefweb-options.js
attiks: Fixed, attributes are copied and `Drupal.attachBehaviors(popup)` is called

orakili (changes requested):
Sorry, there are still issues when selecting options in the popup.

For example, in the training form, if I select 3 themes in the popup and close it, the other options in the original field are not disabled which allows to select one more option (before the behavior kicks in and disables the remaining options).

On the **job** form, selecting "Administration/Finance" for example in the popup doesn't disable the theme field options as it should.

orakili (approved):
It seems to be working now, thanks. I'm approving but this needs to be styled before merging.

Also, for info, while testing I came across this issue: https://humanitarian.atlassian.net/browse/RW-605 which also happens on prod so it's related to this new feature.

orakili:
@attiks For info, the fix in RW-605 (`document` -> `context`) broke the selection limit in the popup:

```
var elements = context.querySelectorAll('fieldset[data-with-selection-limit]:not([data-with-selection-limit-processed])');
```

because `context` is the `popup` element when calling `Drupal.attachBehaviors(popup)` and `popup` was the element with the `data-with-selection-limit` attribute.

So in https://github.com/UN-OCHA/rwint9-site/pull/389/commits/5267f5d38e434d0130ca570732ed08e15407d563, I added the attributes to the `container` element in the popup instead, to fix the issue.

I also changed back the `popup` to a `div` because it's not necessary to have it as a `fieldset` anymore and that will make it easier to style.

cc @left23 just a tiny ping to let you know that the popup most outer element is now a `div` instead of a `fieldset` like it is for the guidelines popups so it should be easier to re-use the styling from the `common_design_subtheme/components/rw-guidelines` component.

orakili (dismissed):
Just re-approving as the last commit dismissed my previous review.

html/modules/custom/reliefweb_form/js/widget.selectionlimit.js
```diff
@@ -71,7 +71,7 @@
       }
 
       // Enable selection limit on relevant checkboxes fields.
-      var elements = context.querySelectorAll('fieldset[data-with-selection-limit]:not([data-with-selection-limit-processed])');
+      var elements = context.querySelectorAll('[data-with-selection-limit]:not([data-with-selection-limit-processed])');
```
attiks, line 74: This means it will bind to all individual input elements as well, I just tried with the previous code and limit is working in page and popup :confused:

orakili: Ah, I didn't notice that the `data-with-selection-limit` was on the individual input as well as the checkboxes container. That was not the case in RW7 and that's also what the `checkSelectionLimit` function implies with `var inputs = element.getElementsByTagName('input');`.

Created https://humanitarian.atlassian.net/browse/RW-611 to solve the problem at the root and only have the attribute on the checkboxes container and not the individual inputs.

orakili (approved):

PR #393, opened 2022-07-01
bug: check disabled

attiks:
Refs: #RW-605


orakili (approved):
That seems to do the trick. Thanks.

PR #396, opened 2022-07-05
[RW-608] Optional mark

orakili:
Refs: RW-608

This fixes an issue where the "optional" mark is not removed for fields that become required via drupal.states.

Ex: the headline title and summary fields on the report form when the headline checkbox is checked.

**Test**

1. Go to the `/node/add/report` form
2. Check the `headline` checkbox -> this should remove the "optional" mark on the headline title and summary fields and add it back when unchecking


left23 (approved):
Works as described. Approving.

PR #399, opened 2022-07-05
feat: queue filtered re-indexing for processing later

lazysoundsystem:
Refs: RW-588

Includes a new entity_before_save hook, which was added reluctantly. I had
trouble getting the unchanged content of the entity in hook_entity_update or
hook_entity_after_save.

### DIFF
--- a/html/modules/custom/reliefweb_api/reliefweb_api.module
+++ b/html/modules/custom/reliefweb_api/reliefweb_api.module
@@ -12,19 +12,12 @@ use RWAPIIndexer\Bundles;
 use RWAPIIndexer\Manager;
 
 /**
- * Implements hook_entity_before_save().
- *
- * @see \Drupal\reliefweb_entities\BundleEntityStorageInterface
+ * Implements hook_ENTITY_TYPE_update().
  */
-function reliefweb_api_entity_before_save(EntityInterface $entity) {
-  // Check if we need re-indexing.
-  if (!$entity->isNew() && $entity->getEntityTypeId() == 'taxonomy_term') {
-    $bundle = $entity->bundle();
-    // Only these types get checked for automatic re-indexing.
-    $taxonomy_types = ['country', 'disaster', 'source'];
-    if (in_array($bundle, $taxonomy_types)) {
-      reliefweb_api_check_reindexing($entity, $bundle);
-    }
+function reliefweb_api_taxonomy_term_update(EntityInterface $entity) {
+  // Check if term needs re-indexing.
+  if (!$entity->isNew()) {
+    reliefweb_api_check_reindexing($entity);
   }
 }
 
@@ -37,6 +30,13 @@ function reliefweb_api_entity_after_save(EntityInterface $entity) {
   // @todo remove when removing `reliefweb_migrate`.
   if (!empty($entity->_is_migrating)) {
     return;
+  }
+
+  if (!empty($entity->needs_reindex)) {
+    $reindex_queue = \Drupal::state()->get('reliefweb_api.reindex_queue');
+    $reindex_queue[$entity->bundle()][$entity->id()] = $entity->id();
+    \Drupal::state()->set('reliefweb_api.reindex_queue', $reindex_queue);
+    unset($entity->needs_reindex);
   }
 
   reliefweb_api_handle_entity($entity);
@@ -170,19 +170,24 @@ function reliefweb_api_execute(array $options, $entity_type_id) {
  *
  * @param \Drupal\Core\Entity\EntityInterface $entity
  *   Entity that might require re-indexing.
- * @param string $bundle
- *   Taxonomy type of entity.
  */
-function reliefweb_api_check_reindexing(EntityInterface $entity, $bundle) {
-  $needs_reindex = FALSE;
-  // Get unchanged entity for comparison.
-  // Tried loadUnchanged() via hooks _entity_update() and _entity_after_save()
-  // but neither gave the original values.
-  $unchanged_entity = \Drupal::entityTypeManager()
-    ->getStorage($entity->getEntityTypeId())
-    ->loadUnchanged($entity->id());
+function reliefweb_api_check_reindexing(EntityInterface &$entity) {
 
-  switch ($bundle) {
+  if (!isset($entity->original)) {
+    return;
+  }
+  $original_entity = $entity->original;
+
+  switch ($entity->bundle()) {
+    case 'country':
+      $fields = [
+        'name',
+        'field_shortname',
+        'field_iso3',
+        'field_location',
+      ];
+      break;
+
     case 'disaster':
       $fields = [
         'name',
@@ -192,11 +197,17 @@ function reliefweb_api_check_reindexing(EntityInterface $entity, $bundle) {
       ];
       break;
 
-    case 'country':
+    case 'disaster_type':
       $fields = [
         'name',
-        'field_shortname',
-        'field_iso3',
+        'field_disaster_type_code',
+      ];
+      break;
+
+    case 'language':
+      $fields = [
+        'name',
+        'field_language_code',
       ];
       break;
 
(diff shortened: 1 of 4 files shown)


html/modules/custom/reliefweb_api/src/Commands/ReliefWebApiCommands.php
```diff
@@ -300,4 +298,68 @@ public function replace($bundle, $newtag, $oldtag = NULL, array $options = [
...
+        continue;
+      }
+      // Launch the re-indexing.
+      try {
+        $options['filter'] = $bundle . ':' . implode(',', $reindex_queue[$bundle]);
+        if ($bundle == 'country' || $bundle == 'source') {
+          $this->index('training', $options);
+          $this->index('job', $options);
+        }
+        $this->index('report', $options);
```
orakili, line 353: There are more things to re-indexed based on the vocabulary (for example source and disaster when the bundle is "country".).


The method below should return all the bundles that need to be re-indexed for the queued vocabulary:

```php
  /**
   * Get the list of bundles to re-index for the given taxonomy vocabulary.
   *
   * @param string $vocabulary
   *   Vocabulary.
   *
   * @return array
   *   List of bundles to re-index.
   */
  protected static function getBundlesToReindex($vocabulary) {
    // Gather a list of all entity reference fields.
    $entity_reference_fields = \Drupal::service('entity_field.manager')
      ->getFieldMapByFieldType('entity_reference');

    // Generate the list of field config IDs for the reference fields.
    $ids = [];
    foreach ($entity_reference_fields as $entity_type_id => $fields) {
      foreach ($fields as $field_name => $info) {
        foreach ($info['bundles'] as $entity_bundle) {
          $ids[] = "$entity_type_id.$entity_bundle.$field_name";
        }
      }
    }

    // Load the configuration of the entity reference fields.
    $field_configs = \Drupal::service('entity_type.manager')
      ->getStorage('field_config')
      ->loadMultiple($ids);

    // Check if the entity reference fields reference the vocabulary in which
    // case retrieve the list of bundles using those fields. Those bundles are
    // the one to re-index.
    $bundles = [];
    foreach ($field_configs as $field_config) {
      $field_name = $field_config->getName();
      $entity_type_id = $field_config->getTargetEntityTypeId();

      if ($field_config->getSetting('target_type') === 'taxonomy_term') {
        $handler_settings = $field_config->getSetting('handler_settings');
        if (isset($handler_settings['target_bundles'][$vocbulary])) {
          $bundles += $entity_reference_fields[$entity_type_id][$field_name]['bundles'];
        }
      }
    }

    // Exclude bundles that are not indexed in the API.
    return array_keys(array_intersect_key(Bundles::$bundles, $bundles));
  }
  ```
  
 Ideally, this should be converted to a non static method with the `\Drupal::service()` calls replaced with proper dependency injection.

File: html/modules/custom/reliefweb_api/src/Commands/ReliefWebApiCommands.php [hunk 2]
```diff
@@ -300,4 +298,68 @@ public function replace($bundle, $newtag, $oldtag = NULL, array $options = [
...
+      }
+      catch (\Exception $exception) {
+        $this->logger->error('(' . $exception->getCode() . ') ' . $exception->getMessage());
```
orakili, line 356: Not sure if that's necessary as `::index()` already wraps the indexing into a try/catch.

html/modules/custom/reliefweb_api/src/Commands/ReliefWebApiCommands.php
```diff
@@ -300,4 +298,68 @@ public function replace($bundle, $newtag, $oldtag = NULL, array $options = [
...
+      }
+      $reindex_queue[$bundle] = [];
+    }
+
+    $this->state->set('reliefweb_api.reindex_queue', json_encode($reindex_queue));
```
orakili, line 361:
```suggestion
    }
```

To go with the suggestion above that empties the queue as soon as it's retrieved.

html/modules/custom/reliefweb_api/reliefweb_api.module
```diff
@@ -145,6 +163,70 @@ function reliefweb_api_execute(array $options, $entity_type_id) {
...
+        'field_disaster_type',
+        'moderation_status',
+      ];
+      break;
+
+    case 'country':
+      $fields = [
+        'name',
+        'field_shortname',
+        'field_iso3',
```
orakili, line 199:
```suggestion
        'field_iso3',
        'field_location',
```

orakili (changes requested):

html/modules/custom/reliefweb_api/reliefweb_api.module
```diff
@@ -145,6 +163,70 @@ function reliefweb_api_execute(array $options, $entity_type_id) {
...
+  }
+
+  foreach ($fields as $field) {
+    if ($entity->get($field)->getValue() != $unchanged_entity->get($field)->getValue()) {
+      $needs_reindex = TRUE;
+      break;
+    }
+  }
+
+  if ($needs_reindex) {
```
orakili, line 223: Adding to the queue should be done in the `reliefweb_api_entity_after_save` to ensure the new data is in the database when the reindexing kicks in.

So in `reliefweb_api_check_reindexing` we should add a flag to the entity object:

```php
  if ($needs_reindex) {
    $entity->needs_reindex = TRUE;
  }
```

And in the `reliefweb_api_entity_after_save`:

```php
  if (!empty($entity->needs_reindex)) {
    $reindex_queue = \Drupal::state()->get('reliefweb_api.reindex_queue');
    $reindex_queue[$entity->bundle()][$entity->id()] = $entity->id();
    \Drupal::state()->set('reliefweb_api.reindex_queue', $reindex_queue);
  }
```

orakili (changes requested):

lazysoundsystem:
Thanks for the helpful and clear comments - not sure how I missed the presence of `$entity->original` in `hook_entity_update()`, and the drush command is much improved.

html/modules/custom/reliefweb_api/reliefweb_api.module
```diff
@@ -171,7 +171,7 @@ function reliefweb_api_execute(array $options, $entity_type_id) {
  * @param \Drupal\Core\Entity\EntityInterface $entity
  *   Entity that might require re-indexing.
  */
-function reliefweb_api_check_reindexing(EntityInterface &$entity) {
+function reliefweb_api_check_reindexing(EntityInterface $entity) {
```
orakili, line 174: @lazysoundsystem FYI, the `&` is unnecessary because `$entity` already holds an identifier pointing at the same object as the one in hook_taxonomy_term_update so any `$entity->xxx` here is be done on the object.

See: https://www.php.net/manual/en/language.oop5.references.php

html/modules/custom/reliefweb_api/src/Commands/ReliefWebApiCommands.php
```diff
@@ -57,8 +57,8 @@ class ReliefWebApiCommands extends DrushCommands {
    */
   public function __construct(
     ConfigFactoryInterface $config_factory,
-    EntityFieldManager $entity_field_manager,
-    EntityTypeManager $entity_type_manager,
+    EntityFieldManagerInterface $entity_field_manager,
+    EntityTypeManagerInterface $entity_type_manager,
```
orakili, line 61: @lazysoundsystem FYI, when injecting dependencies, the interface (if it exists, there are few exceptions in Drupal) should be used as type hint of the parameters so that the services can be overridden.

For example, the `web profiler` module replaces some services to be able to perform its profiling. Those services don't extend the Drupal service classes but implements their interfaces. In that case PHP will throw an exception if the interface is not used as type hint.

html/modules/custom/reliefweb_api/reliefweb_api.module
```diff
@@ -230,7 +230,7 @@ function reliefweb_api_check_reindexing(EntityInterface $entity) {
   }
 
   foreach ($fields as $field) {
-    if ($entity->get($field)->getValue() != $original_entity->get($field)->getValue()) {
+    if (!$entity->get($field)->equals($original_entity->get($field))) {
```
orakili, line 233: The direct comparison `$entity->get($field)->getValue() != $original_entity->get($field)->getValue()` was causing any source being saved even without changes to be queued for re-indexing because the homepage value array in the entity was missing an "attributes" property (because it's not used) that was present in the `original` entity.

(Basically the `$entity` fields contain only the necessary properties without the computed ones or the empty ones.)

orakili (approved):
Thanks, it seems to be working well.

PR #414, opened 2022-07-14
[RW-621] Disable form submission during ajax requests

orakili:
Refs: RW-621

This adds a script added when `ajax.js` is added to a form for example, that disables the form submit buttons while an ajax request is being performed (ex: file/image upload) to prevent mistakes.

Note: this could be added to the CD's subtheme with the following commented out in the `common_design_subtheme.info.yml`

```yaml
  # Core ajax.
  core/drupal.ajax:
    - common_design_subtheme/cd-ajax-disable-submit-buttons
```

so that devs could enable this behavior if interested.

## Test

1. In your browser of choice, if the option is available, set some throttling so that you can see the effect (ex: Firefox dev tools > Network tab > Throttling -> Regular 3G)
2. Go to the `report` form go the "Files" section, upload a large enough file (ex: 500Kb if throttling is enabled)
3. While the file is being uploaded, go to "Save" section and check that the buttons are disabled
4. When the upload is finished, check that the buttons are enabled

To test that the buttons are also properly re-enabled when an ajax request fails, select a file with an unsupported file extension (i.e. not in `csv doc docx jpg jpeg odp ods odt pdf png pps ppt pptx svg xls xlsx zip` and follow the step above. The upload will fail (ajax error) but the buttons will be re-enabled.

### DIFF
--- a/html/themes/custom/common_design_subtheme/common_design_subtheme.info.yml
+++ b/html/themes/custom/common_design_subtheme/common_design_subtheme.info.yml
@@ -59,6 +59,10 @@ libraries-extend:
   reliefweb-guidelines/reliefweb-guidelines:
     - common_design_subtheme/rw-guidelines
 
+  # Core ajax.
+  core/drupal.ajax:
+    - common_design_subtheme/cd-ajax-disable-submit-buttons
+
 # Use the common_design_subtheme styles inside the editor.
 ckeditor_stylesheets:
   - css/styles.css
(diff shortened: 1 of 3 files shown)


orakili:
@rupl  Pinging you to see if you think that can be an interesting addition to the CD.

attiks (approved):

left23 (approved):

PR #416, opened 2022-07-15
[RW-624] Guideline preview

orakili:
Reds: RW-624

This PR updates the GuidelineForm class override to handle the preview added in https://www.drupal.org/project/guidelines/releases/1.0.4 and style the preview to be consistent with the node and term preview on RW.

### Test

Run `composer install` after checking out the branch to update the `guidelines` module.

1. Open a guideline page
2. Edit it, make some changes
3. Click Preview, check that the preview looks like the guideline page from (1) but with the changes
4. Click a link and check that there is a confirmation popup
5. Click "back to content editing"
6. Check that changes from (2) are still present 
7. Save and check that the changes were applied

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "687ff3598a1528e7325a51c23477da0a",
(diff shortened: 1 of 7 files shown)


attiks:
FYI: https://www.drupal.org/project/guidelines/releases/1.0.4

left23 (approved):
Works as described.

PR #419, opened 2022-07-20
[RW-513] Message for 403 on node page

orakili:
Refs: RW-513

This adds back the display of a message for node pages when access is denied to indicate the node is no longer available.

This takes the assumption that people getting a 403 is because the page previously accessible like a published job but is no longer available because the job has expired. The displayed message makes the 403 page a bit more user friendly:

Screen Shot 2022-07-20 at 10 59 00

### Test

1. Log in and find or create a non published (ex: expired) job, training or report, copy the URL
2. Open an incognito tab for example and paste the URL and confirm that the message "The XXX blabla is not available anymore" appears.
3. View another page anonymously like`/moderation/content/report` and confirm there no "not available ..." message on the 403 page.

### DIFF
--- a/html/modules/custom/reliefweb_entities/reliefweb_entities.module
+++ b/html/modules/custom/reliefweb_entities/reliefweb_entities.module
@@ -14,11 +14,13 @@ use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Render\Element;
+use Drupal\Core\Url;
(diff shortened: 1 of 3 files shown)


left23 (approved):
It works as described for an expired job.
But I'm not sure what I should be seeing for Reports.
I get this:
Selection_012
For a Report that is on-hold 
https://rwint.test/report/syrian-arab-republic/ngos-warn-humanitarian-catastrophe-should-un-security-council-fail-allow-aid-be-delivered-syria

orakili:
@left23 Thanks. The screenshot shows the correct result. URLs are only shared when published so if someone gets a 403 it probably means they found the URL somewhere when then node was published (like in a notification email) in which case the message is correct.

PR #422, opened 2022-07-20
[RW-389] Add search converter to RW

orakili:
Refs: RW-389

At the bottom of the rivers like https://reliefweb.int/updates there is an `API` link that redirects to the RW search converter that converts the river URL to an API query which can be used to get data similar to what is displayed in the river.

This search converter is currently a github page: https://reliefweb.github.io/search-converter.

This PR adds a search converter page directly on the RW site so that we don't have to duplicate the conversion logic which is already used for the rivers: `/search/converter` which is accessible via the API link at the bottom of the rivers.

### Test

Checkout this branch and run `drush cr` and `drush cim`.

1. Go the https://reliefweb.int/updates, select some filters and add a full text search query
2. Click the `API` link at the bottom to go the github page search converter
3. Go to your local RW9 site, select the same filters and search query
4. Click the `API` link at the bottom to go the search converter page on RW
5. Compare the `query string`, `api url` and `payload`, they should be identical between both search converter pages

Note: there may be some slight differences if you select some weird combination of filters because the github page is a bit outdated.

### DIFF
--- a/config/reliefweb_rivers.settings.yml
+++ b/config/reliefweb_rivers.settings.yml
@@ -1,1 +1,1 @@
-search_converter_url: 'https://reliefweb.github.io/search-converter'
+search_converter_url: ''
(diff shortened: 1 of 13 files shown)


left23 (approved):
Works as described.

PR #427, opened 2022-07-22
[RW-274] View of revisions

orakili:
Refs: RW-274

This fixes an issue with the entity title that preventing viewing a revision.

It also contains a fix to display private files from old revisions for people with the proper permission.

### Test

**Node revision**

1. Create a report with an attachment and save
2. Edit the report, remove the file, change title, save
3. Check that the file doesn't appear anymore
4. Go to the "revisions" tab for the report, click the first revision
5. Check that this displays the file and the original title

**Term revision**

1. Create a disaster or source and save
2. Edit the term and change the title
3. Go to the "revisions" tab for the term, click the first revision
4. Check that this displays the original title


left23 (approved):
Works are described.

PR #428, opened 2022-07-27
[RW-564] view attached images when previewing node

orakili:
Refs: RW-564

This adds a workaround to be able to view the images when previewing a node with fields using an inline entity form for those image (ex: report image).

### Test

1. Create a report, add an image (under the Files section), fill the mandatory fields
2. Click preview and confirm you can see the image
3. Go back to the form and save
4. Edit the report, edit the image (change the uploaded image for example), click "update image"
5. Click preview and confirm you can see the new image


left23 (approved):
Works as described.

PR #429, opened 2022-07-27
[RW-629] ensure media and files are deleted when deleting a node

orakili:
Refs: RW-629

This ensures the attached media images and their files are deleted when deleting a report or announcement.

### Test

1. Create a node, attach an image (under Files section), save
2. Run `drush sql-query "SELECT * FROM file_managed ORDER BY fid DESC LIMIT 1"` and check the status of  the file is `1`. Note the `fid` and replace `FID` with it in the queries in (4) and (6) below.
3. Delete the node
4. Run `drush sql-query "SELECT * FROM file_managed WHERE fid = FID"` and check the status of  the file is `0`.
5. Run `drush cron`
6.  Run `drush sql-query "SELECT * FROM file_managed WHERE fid = FID"` and check that there is no record anymore.

### DIFF
--- a/html/modules/custom/reliefweb_entities/reliefweb_entities.module
+++ b/html/modules/custom/reliefweb_entities/reliefweb_entities.module
@@ -15,6 +15,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Url;
+use Drupal\media\MediaInterface;
 use Drupal\reliefweb_entities\BundleEntityInterface;
 use Drupal\reliefweb_entities\BundleEntityStorageInterface;
 use Drupal\reliefweb_entities\EntityFormAlterServiceBase;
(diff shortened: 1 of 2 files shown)


lazysoundsystem (approved):
Approving, but I didn't find what you expected in stage 4: on deleting the node through the interface, the file also disappeared from the file_managed table. I didn't run cron as it was already gone.

orakili:
@lazysoundsystem Did you test the image field on the report form?
Screen Shot 2022-07-28 at 8 11 38

lazysoundsystem:
Yes, you're right, I was uploading a file not an image.

Uploading an image, I get the status change in step 4 as described, but running cron (a number of times, each of which takes a long time) doesn't remove the file from the file_managed table.

I'll try refreshing the database and testing again.

lazysoundsystem:
Still the same place - I refreshed the database a number of times, and cron takes an age to run each time: I'm getting the status of the file changed to '0' (except when I'd uploaded the file to more than one node, which I think is expected), but the file remains in the `file_managed` table after cron has run.

This may well be to do with my local set up as there are still some issues. If you're finding this works for you, I'm happy to leave my review as 'approved'.

orakili:
@lazysoundsystem Ah, yes, forgot to mention to add something like `$config['system.file']['temporary_maximum_age'] = 1;` to your local settings.php file. The default is 6h, that's why cron didn't delete the temporary file. The fact that you see the status changed to `0` is all that matters. Thanks for testing again. Merging.

PR #431, opened 2022-07-28
[RW-630] Nginx rule for derivative images

orakili:
Refs: RW-630

This adds a nginx rule for **public** derivative images to return a 404 for unrecognized URLs (ex: unrecognized image style) and otherwise, if the derivative image has not yet been generated, check that the source image exists before passing the request to Drupal.

### DIFF
--- a/docker/etc/nginx/custom/03_derivative_images.conf
+++ b/docker/etc/nginx/custom/03_derivative_images.conf
@@ -1,0 +1,38 @@
+## Location for derivative images to avoid hitting Drupal for invalid URLs or
+## if the source image doesn't exist.
+location /sites/default/files/styles/ {
+
+  ## Allowed derivative images.
+  ## As mentioned in the main README, this needs to be updated when
+  ## new image styles or different directories to store images are added.
+  location ~ "^/sites/default/files/styles/(announcement|icon|thumbnail|small|medium|large|media_library|guideline_thumbnail)/public/(?<file_path>(images|previews)/[^?]+)" {
+    ## The following lines are copied from the apps/drupal/drupal.conf file.
+    access_log off;
+    expires 30d;
+    ## No need to bleed constant updates. Send the all shebang in one
+    ## fell swoop.
+    tcp_nodelay off;
+    ## Set the OS file cache.
+    open_file_cache max=3000 inactive=120s;
+    open_file_cache_valid 45s;
+    open_file_cache_min_uses 2;
+    open_file_cache_errors off;
+
+    try_files $uri @drupal-generate-derivative-image;
+  }
+
+  ## Simply return a 404 for unrecognized derivative URLs.
+  return 404;
+}
+
+## Internal location to ask Drupal to generate a derivative image if the source
+## image is present.
+location @drupal-generate-derivative-image {
+  ## If the source image doesn't exist return a 404.
+  if (!-f "$document_root/sites/default/files/$file_path") {
+    return 404;
+  }
+
+  ## Otherwise pass the request to drupal.
+  try_files /dev/null @drupal;
+}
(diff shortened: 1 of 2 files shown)


docker/etc/nginx/custom/03_derivative_images.conf
```diff
@@ -0,0 +1,38 @@
...
+    access_log off;
+    expires 30d;
+    ## No need to bleed constant updates. Send the all shebang in one
+    ## fell swoop.
+    tcp_nodelay off;
+    ## Set the OS file cache.
+    open_file_cache max=3000 inactive=120s;
+    open_file_cache_valid 45s;
+    open_file_cache_min_uses 2;
+    open_file_cache_errors off;
```
orakili, line 19: @cafuego This was copied from the base drupal.conf: https://github.com/UN-OCHA/docker-images/blob/294f6d052a8d8d4d4505765c466bee56bac9e9d1/php/php-k8s-v8/etc/nginx/apps/drupal/drupal.conf#L145-L158. Do you think that's still relevant as is?

cafuego: I think we can clean that up; but that's low priority. Most of those asset files would be getting cached by varnish anyway, so performance tweaks in nginx are ... not very helpful :-)

docker/etc/nginx/custom/03_derivative_images.conf
```diff
@@ -0,0 +1,38 @@
+## Location for derivative images to avoid hitting Drupal for invalid URLs or
+## if the source image doesn't exist.
+location /sites/default/files/styles/ {
+
+  ## Allowed derivative images.
+  ## As mentioned in the main README, this needs to be updated when
+  ## new image styles or different directories to store images are added.
+  location ~ "^/sites/default/files/styles/(announcement|icon|thumbnail|small|medium|large|media_library|guideline_thumbnail)/public/(?<file_path>(images|previews)/[^?]+)" {
```
cafuego, line 8: I am tempted to rewrite this (for any sub-directory under `styles/`) so we can ship it in the base php-k8s image and have all Drupals take advantage of it. Apart from the directory names, this is not RW-specific.

orakili: @cafuegor Yep, I was thinking about that but then that means we lose the implied image style validation and it still becomes possible to hammer Drupal with unwanted requests.

That aside, the rule needs some testing regarding what nginx captures in the `file_path` variable.

On RW9, the files are stored with their UUID so there is no encoding issue but I remember having issues on RW7 with files containing a `#` character for example and so I didn't add such a rule for it.

I will do some tests with various files and let you know how that went.

orakili: Tested with files containing spaces, `#`, `?` and `%` and that seems to work well. The `$file_path` variable contains the proper url decoded file path.

Created https://github.com/UN-OCHA/docker-images/pull/306 to add the rule to the `php-k8s` images.

cafuego: It would be possible to hammer *nginx* with unwanted requests, but those would never end up hitting Drupal because the validation rule for the source file still need to pass before the request is handed off to Drupal.

orakili: For example `https://reliefweb.int/sites/default/files/styles/UNKNOWN-IMAGE-STYLE/public/images/reports/f8/cd/f8cdb321-08ff-4b18-9d1e-bc6c608c30f8.png` would be passed to Drupal without the whitelist of image styles because the `/images/reports/f8/cd/f8cdb321-08ff-4b18-9d1e-bc6c608c30f8.png` exists and would result in a Drupal 404 (because the image style doesn't exist).

That said, I think the general rule from https://github.com/UN-OCHA/docker-images/pull/306 is good enough for now as it handles the cases I saw in the logs which were for attachment previews or images of reports that were unpublished or deleted.

So closing this in favor of https://github.com/UN-OCHA/docker-images/pull/306.

cafuego:
If it comes to people guessing image presets (to exploit I guess?) by passing actual filenames I guess we'll see an uptick in logs again.

PR #433, opened 2022-07-28
[RW-632] Encode file name when doing legacy redirections

orakili:
Refs: RW-632

Found some occasional nginx 500 in the logs for legacy attachment URLs. It turned out those URLs contained bytes like `\x1` that are considered unsafe when redirecting.

This PR makes sure the file name is encoded before redirecting.

### Test

1. Before switching to this PR's branch, check this URL: `/sites/reliefweb.int/files/resources/ED1EF744FE93A788C%1F1257428003110CB-gvtSweden_feb2006.pdf`, it should return a 500
2. Checkout this PR's branch, **restart the container** and check the same URL, it should return a 404


lazysoundsystem (approved):
Just restarting the container wasn't enough, but when I rebuilt it from scratch, it worked as described.

PR #434, opened 2022-07-28
[RW-633] skip 403 preprocessing if the router throws an exception

orakili:
Refs: RW-633

This fixes an issue where Drupal returns a 500 instead of a 403 when selecting a source in a job/training form as a non-editor.

Note: code is best reviewed with `?w=1` as all this does is wrapping the code in a try/catch.

### Tests

Before checking out this branch:

1. Go to `/node/add/job` as non Editor
2. Open your browser inspector's network tab
3. Select a source (ex: OCHA) and check that there is a call to `/admin/reliefweb_form/node_form/source_attention_messages/job` that returns a 500

Checkout the branch, do the same steps as above but this time, the return status should be 403 (expected as the user doesn't have the right to access the source_attention_messages endpoint).


left23 (approved):
Works as described

PR #435, opened 2022-07-29
[RW-636] Preload SVG and font stylesheet

orakili:
Refs: RW-636

This adds `<link>` elements to the head to preload the SVG sprites and logos and the google font stylesheet to reduce the time to the first contentful paint.

**Before**
Screen Shot 2022-07-29 at 10 39 54

**After**
Screen Shot 2022-07-29 at 10 40 06


rupl (approved):
Great improvement! I'd be careful not to load too many sprites up front as they could jump the line in front of the CSS that calls them, but the fonts are definitely a good thing to preload/preconnect. Especially on a cold browser cache it will make a big difference.

PR #438, opened 2022-08-05
RW-647 Mailchimp form accessibility fixes

left23:
Refs: RW-647


orakili (dismissed):
Looks good to me. Just wondering if we should not change the honeypot label to something like "Do not fill in this field" as that may be more understandable for non tech people.

left23:
> Looks good to me. Just wondering if we should not change the honeypot label to something like "Do not fill in this field" as that may be more understandable for non tech people.

Yes, much better :)

orakili (approved):

PR #443, opened 2022-08-12
[RW-654] Allow topics to be archived

orakili:
Refs: RW-654

This simply enables the "archive" status for topics.

### Tests

1. Select a published topic, check that it can be accessed as anonymous
2. Log in as an Editor
3. Edit the topic, check that there is an "Archive" button, click on it
4. Check that the topic status is "archived" and that the topic is not accessible anonymously anymore5. 
5. Check that, as an Editor the topic is accessible and editable
6. Log in as a normal user and check that the topic is not accessible


left23 (approved):
Works as described!

PR #449, opened 2022-09-01
[RW-658] Ensure source status field is properly updated

orakili:
Refs: RW-658

This updates the `SourceModeration` service to call the parent `entityPresave` to ensure the term `status` field is properly updated when the moderation status changes.

### Tests

1. Find a source with a "status" of 0: `SELECT tid FROM taxonomy_term_field_data WHERE status = 0 LIMIT 1`
2. Edit the source, save it as `active`
3. Check that the status is still `0`: `SELECT status FROM taxonomy_term_field_data WHERE tid = TID` (replace `TID`)

1. Find a source with a "status" of 0: `SELECT tid FROM taxonomy_term_field_data WHERE status = 0 LIMIT 1`
2. Edit the source, save it as `active`
3. Check that the status is now `1`: `SELECT status FROM taxonomy_term_field_data WHERE tid = TID` (replace `TID`)
4. Edit the source, save it as `archive`
5. Check that the status is now `0`: `SELECT status FROM taxonomy_term_field_data WHERE tid = TID` (replace `TID`)

### DIFF
--- a/html/modules/custom/reliefweb_moderation/src/Services/SourceModeration.php
+++ b/html/modules/custom/reliefweb_moderation/src/Services/SourceModeration.php
@@ -208,6 +208,8 @@ class SourceModeration extends ModerationServiceBase {
    * {@inheritdoc}
    */
   public function entityPresave(EntityModeratedInterface $entity) {
+    parent::entityPresave($entity);
+
     // Ensure all posting rights are 'blocked' if the status is 'blocked'.
     $status = $entity->getModerationStatus();
     if ($status === 'blocked') {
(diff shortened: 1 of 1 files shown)


orakili:
@lazysoundsystem FYI, filters for unpublished terms are removed when sanitizing the advanced-search query parameter: https://github.com/UN-OCHA/rwint9-site/blob/f5fbc6e9e3ceb6e97f132040621616faadd1473c/html/modules/custom/reliefweb_rivers/src/AdvancedSearch.php#L1002.

The issue was that the `status` field for the PMM source always stayed at `0` regardless of the moderation status (active, inactive) because of the missing call to the parent method added in this PR.

That resulted in the disappearance of the filter in the jobs river.

PR #450, opened 2022-09-01
fix: Strip quotes, backticks and dollars from the string we're looking up.

cafuego:
Since io.popen() spawns a shell, we do not want to pass it any special
characters that would end up with a shell exploit.

Refs: OPS-8576


orakili (approved):
Looks good to me. Thanks.

PR #453, opened 2022-09-02
[RW-659] Use collation field to sort organizations

orakili:
Refs: RW-659

This uses the collation field introduced in https://github.com/UN-OCHA/rwint-api-indexer/pull/42 to sort the organizations alphabetically.

**Important**

This requires the update of the API indexer in https://github.com/UN-OCHA/rwint-api-indexer/pull/42 and the corresponding update of the API: https://github.com/UN-OCHA/rwint-api/pull/102

This uses the temporary `v2.5.0-alpha` tag of the API indexer for the purpose of testing and that will need to be changed to v2.5.0 when available.

### Tests

**Prerequisites**

1. Install the `analysis-icu` plugin in the elasticsearch container: `docker exec -it rwint9-elasticsearch bin/elasticsearch-plugin install analysis-icu`
2. Update the local API with https://github.com/UN-OCHA/rwint-api/pull/102

1. Go to your local RW site organizations page and click on the `I` letter filter
2. Check that `iMMAP` is not on the first page
3. Click "last" in the page and check that iMMAP is towards the end of the page

1. Run `composer install` to switch to the `v2.5.0-alpha` version of the api-indexer
2. Re-index the sources: `drush rapi-i --tag=rw-659-test1 --replace --alias source`
3. Clear the cache `drush cr`
4. Go to your local RW site organizations page and click on the `I` letter filter
5. Check that `iMMAP` is on the first page and correctly placed among other sources alphabetically speaking

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
         "league/html-to-markdown": "^5.0",
         "lolli42/finediff": "^1.0",
         "pelago/emogrifier": "^6.0",
-        "reliefweb/api-indexer": "^v2.3",
+        "reliefweb/api-indexer": "v2.5.0-alpha",
         "reliefweb/simple-autocomplete": "^v1.3",
         "reliefweb/simple-datepicker": "^v1.3",
         "symfony/flex": "^2.0",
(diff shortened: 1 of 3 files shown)


lazysoundsystem (approved):
Got tied up with an had an unrelated issue of elasticsearch refusing to index anything (fixed with the solution here: https://stackoverflow.com/a/60720259 ). Then installing the analysis-icu plugin required a restart, which removed the flood fix. 

Got everything lined up eventually, and it all works as expected.

PR #455, opened 2022-09-07
[RW-667] Performance improvements

orakili:
Refs: RW-667

This adds a series of performance improvements:

- RW-660: add loading attribute for river thumbnails
- RW-661:  load the announcement image and first headline image eagerly on the homepage
- RW-662, RW-663:  load the report image or first attachment preview eagerly
- RW-664: load blog image eagerly
- RW-665: inline GTM loading snippet
- RW-666: preload mobile logo
- RW-668: use OCHA logo sprite for OCHA logo in footer

### Tests

Run some performance profiling tools (with throttling) on your local site (ex: lighthouse), and record the results for:
- (1) homepage, 
- (2) report with an image, 
- (3) report with an attachment with a preview but no image
- (4) report with an image and an attachment
- (5) blog post with an image

1. Check that the "loading" attribute is set to "lazy" on the thumbnails of the `/updates?view=maps` for example
2. Check that the "loading" attribute is set to "eager" for the first headline image and the announcement on the homepage (1)
3. Check that the "loading" attribute is set to "eager" on the report (2)'s report image
4. Check that the "loading" attribute is set to "eager" on the report (3)'s attachment preview
5. Check that the "loading" attribute is set to "eager" on the report (4)'s image and set to "lazy" on the attachment preview.
6. Check that the "loading" attribute is set to "eager" on the blog post (5)'s report image
7. Check that the GTM loading snippet is inline in the source of a page
8. Check that there is a meta element to preload the rw-logo-mobile.svg file
9. Check your network tab and ensure `ocha-lockup.svg` is not loaded and that the OCHA logo in the footer is present

Run the same performance tool on the same pages as above and compare the results to confirm there is indeed some improvements. This should notably reduce the "LCP" (Largest Contentful Paint).

### DIFF
--- a/config/google_tag.settings.yml
+++ b/config/google_tag.settings.yml
@@ -2,7 +2,7 @@ _core:
   default_config_hash: 65d5DfK5p2hQDpBskbvCIj2hSEqw_Odm2e_d46BH1Lk
 uri: 'public:/'
 compact_snippet: true
-include_file: true
+include_file: false
 rebuild_snippets: true
 flush_snippets: false
 debug_output: false
--- a/html/modules/custom/reliefweb_entities/reliefweb_entities.module
+++ b/html/modules/custom/reliefweb_entities/reliefweb_entities.module
@@ -297,10 +297,13 @@ function reliefweb_entities_theme() {
         'attributes' => NULL,
         // Image style.
         'style' => NULL,
-        // Image information with uri, width, height, alt and copyright.
+        // Image information with uri, width, height, alt and copyright
+        // and optionally the loading mode.
         'image' => [],
         // Flag to indicate whether to show the caption or not.
         'caption' => TRUE,
+        // Flag to indicate the default loading mode: lazy or eager.
+        'loading' => 'lazy',
       ],
     ],
     'reliefweb_entities_book_menu' => [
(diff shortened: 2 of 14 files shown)


lazysoundsystem (dismissed):
All works as expected, though I didn't see *a lot* of change in the profiling. This might be because I didn't find the 'throttling' setting until after I'd done the 'before' metrics. 

I did see the numbers go down for 'scripting', 'rendering' and 'painting' and the 'idle' time went up, but the LCP measurements seemed pretty similar.

Just noting that the GTM script includes 'xxx's: src="https://www.googletagmanager.com/gtm.js?id=GTM-xxxxxx" which presumably get switched out by Ansible at some stage.

lazysoundsystem:
After a discussion about profiling in standup, I tried again using Lighthouse instead of chrome dev tools. The results are fairly similar - on Desktop, a 99% Performance score for both before and after, on Mobile a 90% Performance score both before and after, but now with less warnings:
image

orakili:
@lazysoundsystem Just added https://github.com/UN-OCHA/rwint9-site/pull/455/commits/5a3f9a895fd83cd665a751f79bb089e2857fe4c1 that adds a media attribute for the mobile logo so we only preload it on relevant devices.

lazysoundsystem (approved):

PR #457, opened 2022-09-14
fix: Force the file extension to lowercase for file access, as that is how we store it.

cafuego:
This does not affect the file extension in the content-disposition header, which is the same case as the original upload.

Refs: OPS-8473


cafuego:
Note that I have hot-fixed this on all production web heads.

orakili (approved):
Thanks!

PR #458, opened 2022-09-15
[RW-672] Disaster map accessibility

orakili:
Refs: RW-672, RW-671, RW-646

This improves the accessibility of the disaster map by highlighting the disaster markers on the map and showing the popup when navigating by keyboard.

There is also a tiny fix for the icons of the related disasters where the use of the wrong css variable resulted in a wrong position.

### Tests

1. Checkout this branch, clear the cache
2. Open the /disasters page and use `tab` to navigate until reaching the map
3. Check that this highlight markers on the map and shows the popup.
4. Check that related disasters can also be reached with `tab`
5. Check that pressing `escape` closes the popup and than `tab` jumps to the next disaster

Re. (4), here's a screenshot showing related disasters (you may need to tab through several disasters to find one with related disasters):

Screen Shot 2022-09-15 at 11 21 01

### DIFF
--- a/html/modules/custom/reliefweb_disaster_map/js/disaster-map.js
+++ b/html/modules/custom/reliefweb_disaster_map/js/disaster-map.js
@@ -240,6 +240,13 @@
           }
         });
 
+        // Unset the active marker when pressing escape.
+        container.addEventListener('keydown', function (event) {
+          if (event.keyCode === 27) {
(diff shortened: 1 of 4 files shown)


lazysoundsystem (approved):
Tested and works as described (once I added the mapbox key and token on my local setup).

PR #460, opened 2022-09-15
[RW-635] Add sitemap entry to robots.txt

orakili:
Refs: RW-635

This simply appends back (was in D7) the sitemap location to the robots.txt

### Tests

1. Checkout the branch
2. Build the image with make and rebuild container
3. Check /robots.txt to confirm there is "Sitemap: https://reliefweb.int/sitemap.xml" at the end

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 # Ignore everything except for the files necessary to build the codebase.
 *
+!assets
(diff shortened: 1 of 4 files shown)


lazysoundsystem (approved):
Confirming the line is added as it should be.

PR #461, opened 2022-09-16
[RW-673] Enable GTM on 403/404

orakili:
Refs: RW-673

This remove the setting that preventing the GTM loading script from being injected in 403/404 pages.

### Tests

1. Checkout the branch and import the config (`drush cim`)
2. Type a random non existing URL (404) and check that the inline GTM loading script is there
3. Open closed job for example (403) and check that the GTM loading script is there

### DIFF
--- a/config/google_tag.container.default.yml
+++ b/config/google_tag.container.default.yml
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   module:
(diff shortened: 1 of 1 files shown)


lazysoundsystem (approved):
Checked the script appears on both 403 and 404 pages

PR #462, opened 2022-09-16
[RW-634] do not attempt to fetch source attention message when not an editor

orakili:
Refs: RW-634

This disables trying to fetch the attention message when selecting a source in the job or training form when the current is not an editor. Currently this returns unnecessary 403s.

### Tests

1. Log in as a normal user (non editor)
2. Go the /node/add/training to create a training
3. Open the network tab in your browser's dev tools
4. Select "UN Women" and check that there is a request to `admin/reliefweb_form/node_form/source_attention_messages/training` that returns a 403 and that no attention message appears 

Screen Shot 2022-09-16 at 14 30 17

5. Login in as an editor, repeat the steps below and check that the request above returns a 200 and that the attention message appears in the form6. 

Screen Shot 2022-09-16 at 14 29 34

Repeat the steps below and check that at (4) there is no ajax request when logged in as a normal user and that there is a proper request returning a 200 with the display of the attention message (5) when logged in as an editor


lazysoundsystem (approved):
Confirmed this works as it should

PR #470, opened 2022-09-22
[RW-221] Add tab navigation for bookmarks

orakili:
Refs: RW-221

This fixes a couple of issues with the bookmarks (see first 2 commits) and add some tab navigation.

It's re-using the "views" from the river pages like the `/updates` page and I think the styling is good enough for such a seldom used feature.

### Tests

1. Bookmark some reports, jobs and training
Screen Shot 2022-09-22 at 12 14 14

2. Go to your account >  bookmarks
3. Check that there the tab navigation appears and that the links show the proper bookmarks4. 

Screen Shot 2022-09-22 at 12 10 22


lazysoundsystem (approved):
Works as described.

PR #475, opened 2022-09-29
[RW-681] Ensure that the entity publication status and moderation status are in sync

orakili:
Refs: RW-681

The logic for deciding whether or not to set the drupal status field to 0 or 1 was faulty for the announcement, resulting in the status field to be always 0 preventing published announcements from being displayed on the homepage.

### Tests

1. Create a new announcement, save as draft
2. Check that the announcement page is not accessible to anonymous or normal users but that editors can see it
3. Check that the announcement does **not** appear on the homepage
4. Save the announcement as published
5. Check that the announcement page is not accessible to anonymous or normal users but that editors can see it
6. Check that the announcement appears on the homepage

Ideally the following tests should be conducted for all the other existing main content types (jobs, training, disasters etc.):

1. Edit an entity (ex: report), save it with an unpublished status like "draft", "on-hold" or "archive"
2. Check that the page is not accessible to anonymous or normal users but that editors can see it
3. Run `drush sql-query `SELECT status FROM node_field_data WHERE nid = %nid%` (replacing the `%nid%` with the entity id, adjust if the entity is a term)
4. Check that the returned status is 0
5. Save the entity with the a published status like "to-review", "published" or "active"
6. Check that the page is accessible to everybody
7. Run `drush sql-query `SELECT status FROM node_field_data WHERE nid = %nid%` (replacing the `%nid%` with the entity id, adjust if the entity is a term)
8. Check that the returned status is 1

### DIFF
--- a/html/modules/custom/reliefweb_moderation/src/ModerationServiceBase.php
+++ b/html/modules/custom/reliefweb_moderation/src/ModerationServiceBase.php
@@ -17,7 +17,6 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Link;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Session\AccountProxyInterface;
-use Drupal\Core\Session\AnonymousUserSession;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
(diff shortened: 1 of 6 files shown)


lazysoundsystem (approved):
All tested and works as described.

PR #480, opened 2022-10-07
[RW-528] Add reporting module to send posting stats

orakili:
Refs: RW-528

This adds a custom module with some drush commands to send posting statistics to replace logic in jenkins jobs.


attiks (approved):
Sorry for the delay

PR #494, opened 2022-11-02
[RW-692] Datepicker accessibility

orakili:
Refs: RW-692

This improves the accessibility of the datepicker widget in the advanced search (date filters in `/updates` for example), following the pattern from https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/datepicker-dialog.html.

- Use separate button to display the datepicker (see screenshot below)
- Fix keyboard navigation
- Wrap focus when the datepicker dialog is open (select -> previous year)
- Ensure compatibility with datepicker widget in entity forms to avoid disturbing the editorial teams and other posters...

**Screenshot**

Screen Shot 2022-11-01 at 16 55 28

**Notes**

This includes the update of the datepicker library v1.3.3-beta: https://github.com/UN-OCHA/rwint-simple-datepicker/pull/2 - after review composer will be updated to use the v1.3.3 release of the library.

This has only been tested on recent Safari, Firefox and Chrome on MacOS for now.

There is still room for improvements (section on how to use the datepicker for [example](https://fymmot.github.io/inclusive-dates/)) but that's already a good progress in my opinion.

### Tests

1. Check out the branch and run `composer install` to ensure the v1.3.3-beta of the datepicker library is added.
2. Visit `/updates`, use the keyboard to navigate to the `Original publication date` filter and select date (use arrow keys to select day, try with mouse as well etc.)
3.  Check that entering a date in the input (ex: `From`) results in the datepicket showing the proper entered date when opening it.
4. Open prod for example, go to `/node/add/training`, do the same on local and confirm that the date picker for the date fields behave similarly when typing the date or using the mouse to select the date. (The keyboard navigation should be a bit better on the local with that branch though there is no focus wrapping.)


left23 (approved):
Works as described, nice improvements.

PR #497, opened 2022-11-04
[RW-693] Filter accessibility

orakili:
Refs: RW-693

This enables closing the filter dialog with `Escape` and wraps the focus inside the dialog in the river filters (ex: /updates) to improve the keyboard navigation.

It also fixes an existing issue when pressing escape in the moderation backend.

## Tests

1. Checkout the branch and go to `/updates`
2. Open a filter (ex:  Country)
3. Check that the focus is wrapped (add -> first input etc.)
4. Check that the dialog closes when pressing `Escape` similarly to clicking `Cancel`

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8917,16 +8917,16 @@
         },
         {
             "name": "reliefweb/simple-datepicker",
-            "version": "v1.3.3-beta",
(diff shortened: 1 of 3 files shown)


left23 (approved):
Works as described

PR #502, opened 2022-11-11
[RW-701] Fix record total in moderation pages

orakili:
Refs: RW-701

This adds a `DISTINCT` to the main query of the moderation pages to ensure the counting of the document is correct.

### Tests

1. Create or edit a report, add a dummy revision comment, save.
2. Repeat (1).
3. Go to `/moderation/content/report`, select `reviewed date` filter and today's date (so it catches the modification from (1) and (2)).
4. Add a `title` filter with the title of the node modified in (1)
5. There should be only 1 record in the list but the total should be 2 (or more).

Repeat (3) and (4) but this time the total in (5) should be `1`.


lazysoundsystem (approved):
Works as expected with this change - but I somehow got the correct total before checking out this branch too (editing, not creating, two reports).

PR #505, opened 2022-11-11
[RW-646] Disaster/country description visibility toggler

orakili:
Refs: RW-646

This moves the "show/hide description" toggler for countries and disasters after the description to preserve the keyboard navigation order (visual order was different than source order) to improve accessibility.

It also set the focus on the description title when expanding the description to make it easier for the user to read the description and navigate through it with the keyboard.

When pressing the "hide description", the page scrolls to the description toggler button which provides a consistent experience across browsers while otherwise the user can be sent much lower in the page (ex: Maps) depending on the browser.

For that, this PR intoduces some improvements and changes to the `cd-dropdown.js`:
- New methods to get the toggler from the toggable and vice versa using the `aria-controls` attribute and toggler id. This avoids relying on the `previousElementSibling`.
- Add a new attribute `data-cd-insert-after` to insert the toggler after the toggable element which is useful for "show more/show less" toggles.
- Generate a unique ID for the toggable elements that don't have one to be able to use the `aria-controls` mentioned above.

Those changes should not impact existing sites and so the changes should probably make their way to the base theme after review at some point.

### Tests

1. Go to  `/disaster/ep-2016-000126-nga` for example to see an example of a disaster with a large description.
2. Click the "show full description" button and check that the view scrolls back to the "disaster description" title and keyboard navigation allows to go through the links in the description.
3. Click "hide the description" and check that the view scrolls to the toggler button (on prod, the view jumps to middle of the "Maps and Infographics" or some other place depending on the browser).
4. Use the keyboard navigation and confirm (2, pressing enter instead clicking) and (3)

### DIFF
--- a/html/themes/custom/common_design_subtheme/common_design_subtheme.info.yml
+++ b/html/themes/custom/common_design_subtheme/common_design_subtheme.info.yml
@@ -68,6 +68,9 @@ ckeditor_stylesheets:
   - css/styles.css
 
 libraries-override:
+  common_design/cd-dropdown:
+    js:
+      js/cd-dropdown.js: js/cd-dropdown.js
   guidelines/guidelines-json:
     js:
       components/guidelines-json/guidelines-json.js: false
--- a/html/themes/custom/common_design_subtheme/components/rw-entity-text/rw-entity-text.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-entity-text/rw-entity-text.css
@@ -1,7 +1,3 @@
-.rw-entity-text {
-  display: flex;
-  flex-direction: column;
-}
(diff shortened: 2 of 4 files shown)


left23 (dismissed):
@orakili This looks good and is definitely more user-friendly.
When coming from the Disasters River page initially and clicking on a specific Disaster, the page position is not at the top of the page, but instead right above the "Show Full Description" toggle (happens in Chrome and FF on Ubuntu). I'm not sure if this is intended or if it is something that can be improved.

Selection_008

orakili:
@left23 Thank you for posting this problem! It should be fixed now. The page should scroll to the button only when hiding the full description.

left23 (approved):
Looks good!

PR #513, opened 2022-11-18
[RW-710] Fix validation of query parameters

orakili:
Refs: RW-710

This adds some validation to all the places where we retrieve query parameters to avoid unexpected behaviors (like uncaught exception) when the parameter is invalid (ex `search[]=whatever` instead of `search=whatever`).

### Tests

Compare with prod where the pages should show a "The website encountered an unexpected error. Please try again later.".

1. `/updates?search[]=whatever` (expected: parameter ignored)
2. `/updates?view[]=whatever` (expected: parameter ignored)
3. `/updates?advanced-search[]=whatever` (expected: parameter ignored)
4. `/updates?legacy-river[]=country/afg`  (expected: parameter ignored)
5. `/search/results?search[]=whatever` (expected: parameter ignored)
6. `/country/afg?figures[]=all` (expected: parameter ignored; it's already also ignored on prod)
7. `/organizations?group[]=whatever` (expected: parameter ignored)

### DIFF
--- a/html/modules/custom/reliefweb_entities/src/Entity/Topic.php
+++ b/html/modules/custom/reliefweb_entities/src/Entity/Topic.php
@@ -269,7 +269,7 @@ class Topic extends Node implements BundleEntityInterface, EntityModeratedInterf
     $river = $service->getRiver();
 
     // Get the river view.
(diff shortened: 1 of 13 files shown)


lazysoundsystem (approved):
Confirming this fixes the errors

PR #517, opened 2022-11-25
[RW-712] Fixes for translated forms.

orakili:
Refs: RW-712

This includes 2 fixes to improve compatibility of the forms with translation tools (like Chrome's page translation):

1. Use `<button>` instead of `<input>` for submit buttons in order to preserve the `value` attribute that Drupal uses to identify the clicked button. (For buttons, the text inside `<button>LABEL</button>` is translated while for input of type submit, the value is translated.)
2. Strip the `<font>` elements in the sanitizer, keeping the relevant children instead of removing the elements completely. Chrome page translation for example wraps all the translated text inside `<font style="vertical-align: inherit;">...</font>` for whatever reason.

### Tests

(2) above is already checked by the unit tests so only (1) needs testing:

1. Using Chrome, go to settings > languages > add a language other than English and select it as language to use for the translation (may have to remove English (USA) which is the default one first):
2. Log in local RW site
3. Create a new job (link from dashboard)
4. Right click to show the contextual menu and select “Translate to …”
fill in the form
5. Click “submit” → you should be redirected to the preview

Note: alternatively, just edit the `value` of one of the submit buttons in the job form to some gibberish so simulate the translation of the button text.

Repeat the above but this time (5) should properly submit the job.

### DIFF
--- a/html/modules/custom/reliefweb_utility/src/Helpers/HtmlSanitizer.php
+++ b/html/modules/custom/reliefweb_utility/src/Helpers/HtmlSanitizer.php
@@ -170,6 +170,8 @@ class HtmlSanitizer {
       'tr' => FALSE,
       'sup' => FALSE,
       'sub' => FALSE,
+      'small' => FALSE,
+      'font' => FALSE,
       // No children.
       'img' => TRUE,
     ];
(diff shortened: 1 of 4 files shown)


left23 (approved):
Works as described.

PR #520, opened 2022-11-29
[RW-717] Fix query performance on moderation pages

orakili:
Refs: RW-717

This restricts the fix from RW-701 (`distinct()`) to only apply when there is a filter on a revision table as it's the source of the duplication. This improves the performances for queries which already return unique results.

Note: previous PR with the fix from RW-701: https://github.com/UN-OCHA/rwint9-site/pull/502

### Tests.

1. Check out the latest develop
2. Go to `/moderation/content/report`, check how long it takes to load
3. Add a review date filter for example, apply and take note of the totals

1. Go to `/moderation/content/report`, check how long it takes to load => it should be faster than (2) above
2. Add the same date filter as (2) above, apply and check that the totals are the same.


lazysoundsystem (approved):
Works well - I'm using the reduced database so there's not such a difference in speed, but all looks good.

PR #521, opened 2022-11-29
[RW-705] se selected filters and search for river page description

orakili:
Refs: RW-705

This changes the generic page description of the rivers when there are selected filters to help search engines notably differentiate between those pages.

So for example [this search query](https://reliefweb.int/updates?advanced-search=%28S1275%29.%28DT4642_T4595%29_%28DA20200301-%29_%21%28OT273%29) will have the description:

> Updates with organization: World Health Organization (WHO) or with disaster type: Epidemic and theme: Health and with posting date on reliefweb: after 2020/02/29 and without organization type: Media

instead of the generic when there is no filter or search query:

> Your gateway to all content to date. Search and/or drill down with filters to narrow down the content.

### Tests.

1. Checkout this branch
2. Go to a river (ex: /updates), select some filters and apply
3. Check that the description in the `<head>` matches the selected filters
4. Do a full text search as well and check that it also appears in the description
5. Select a view (ex:  headlines) and check that it also appears in the description (ex: `Updates  (Headlines) ...`)

### DIFF
--- a/html/modules/custom/reliefweb_meta/reliefweb_meta.module
+++ b/html/modules/custom/reliefweb_meta/reliefweb_meta.module
@@ -52,6 +52,15 @@ function reliefweb_meta_metatags_alter(array &$metatags, array &$context) {
     'image' => reliefweb_meta_get_image_url(),
   ];
 
+  if (strpos($route_name, 'reliefweb_rivers.') === 0) {
+    try {
+      $description = \Drupal::service($route_name)->getRiverDescription();
+    }
+    catch (Exception $exception) {
+      // Nothing to do.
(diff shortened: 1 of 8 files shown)


lazysoundsystem (approved):
Works as planned - 
We *could* consider putting some kind of quotes around the search term e.g `<meta name="description" content="Updates containing &quot;test search term&quot; and with organization: World Health Organization (WHO) or with disaster type: Epidemic and theme: Health and with posting date on reliefweb: after 2020/02/29 and without organization type: Media" />`
but that might make it more complicated if someone searches for a term with quotes included.

PR #526, opened 2022-12-02
[RW-718] Fix cache control of RSS feeds and JSON output

orakili:
Refs: RW-718

This ensures the Cache-Control header for the RSS feeds (ex: /updates/rss.xml) and the json output of the search converter is properly set so that varnish can cache the pages.

### Tests

Note: this requires that `$config['system.performance']['cache']['page']['max_age']` is properly set in the `settings.local.yml` file.

1. Check the cache control header of `/updates/rss.xml`, it should be  `private`
2. Check the cache control header of `/search/converter?appname=test&search-url=https%3A//rw9.test/updates`, it should be private

Repeat the above but this time the cache-control header should be `max-age=XX, public` where `XX` is the value of `$config['system.performance']['cache']['page']['max_age']` or `60` if using varnish.

### DIFF
--- a/html/modules/custom/reliefweb_rivers/src/Controller/SearchConverter.php
+++ b/html/modules/custom/reliefweb_rivers/src/Controller/SearchConverter.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\reliefweb_rivers\Controller;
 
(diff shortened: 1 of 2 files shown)


lazysoundsystem (approved):
Yes, works as described.

PR #529, opened 2022-12-13
[RW-723] Logic fix for my posts page for blocked sources

orakili:
Refs: RW-723

This fixes the logic of the my posts page when the user is blocked for a source to display all the user's ads except the ones from sources the user is blocked for.

### Tests.

*Pro tip: log in with the normal user in a normal window and log in with an editor with a private window to avoid logging in and out several times*

1. Log in as a normal user (create a dummy one for example)
2. Create 2 jobs with different sources
3. Log in as an editor, edit the user posting rights for the sources above and set the previous user as allowed for one and blocked for the other source
4. Log in back as the normal user
5. Check your "My Posts" page (/user/xxxx/posts)
6. You should **only** see the ad for the source the user is blocked for
7. Check the user's posts page as an editor and both ads should appear

Repeat (4) to (7) above. In (6), you should **only** see the ad for the source the user is allowed to post for.

### DIFF
--- a/html/modules/custom/reliefweb_user_posts/src/Services/UserPostsService.php
+++ b/html/modules/custom/reliefweb_user_posts/src/Services/UserPostsService.php
@@ -567,10 +567,13 @@ class UserPostsService extends ModerationServiceBase {
     if (!empty($blocked) && !$this->currentUser->hasPermission('edit any job content')) {
       foreach ($types as $type) {
         if (!empty($blocked[$type])) {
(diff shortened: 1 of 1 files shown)


lazysoundsystem (approved):
Sorry for the delay on this - I can confirm it works as described.

Unrelated, but I noticed a different bug along the way: https://humanitarian.atlassian.net/browse/RW-724

PR #530, opened 2022-12-14
[RW-724] Fix posted filter on people page

orakili:
Refs: RW-724

This fixes the exception thrown when using the "posted" filter on the people page.

### Tests

1. Go to `/admin/people`
2. Check "Job" for example under "Posted"
3. Click "Filter"
4. Cry

Repeat the above but this time at (4), after a few seconds (because the query is slow with this filter), you should see the properly filtered list of users.

### DIFF
--- a/html/modules/custom/reliefweb_users/src/Form/UserPageFilterForm.php
+++ b/html/modules/custom/reliefweb_users/src/Form/UserPageFilterForm.php
@@ -96,7 +96,7 @@ class UserPageFilterForm extends FormBase {
       '#title' => $this->t('Posted'),
       '#type' => 'checkboxes',
       '#options' => $content_types,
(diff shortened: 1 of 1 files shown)


lazysoundsystem (commented):
Got a similar error: `TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in array_keys() (line 99 of modules/custom/reliefweb_users/src/Form/UserPageFilterForm.php). `

Changing the `NULL` in line 197
```
'posted' => isset($inputs['posted']) && is_array($inputs['posted']) ? array_intersect_key($content_types, $inputs['posted']) : NULL,
```
to `[]` fixed it for me.

lazysoundsystem (approved):
Adding approval so as not to hold this up when it's fixed.

PR #533, opened 2022-12-27
[RW-725] User routes

orakili:
Refs: RW-725

This adds `/user/*` routes and associated controllers/methods for the bookmarks, posts and subscriptions user pages for the current user so we can add menu items under the new "my account" dropdown added in CD 7.4.

### Tests

1. Checkout this branch and clear the cache (`drush cr`)
2. Log in with a user.
3. Confirm that the `/user/bookmarks`, `/user/notifications` and `/user/posts` redirect to the corresponding `/user/{uid}/*` pages.


lazysoundsystem (approved):
Works as described.

PR #439, opened 2022-08-10
[RW-21] User revisions - DO NOT MERGE YET

orakili:
Refs: RW-21

This adds the user_revision module (MR3 of https://www.drupal.org/project/user_revision/issues/3197068) and adjusts the reliefweb_revisions module to be more flexible and do not assume all revisioned entities are also moderated ones.

### Tests

1. Checkout this branch and run `composer install` to install the user_revision
2. Run `drush cr` then `drush cim`, this should enable the user_revision module and start a batch to make the user entities revisionable (can take several minutes)
3. Log in with a an existing "normal" user, change the display name or email address
4. Log in with an account with the "User Manager" role, edit the account from (3), there should be the revision history at the bottom of the form showing the change of display name/email address
5. Change something for the account (3), leave a revision comment, save and edit again to confirm the changes and check they also appear in the revision history with the comment

**Important:** This makes changes to the DB so either backup your DB before reviewing this PR or import a fresh dump when done with the review and checking out another branch.

PR #543, opened 2023-01-12
[RW-21] Store user change history

orakili:
Refs: RW-21

This replaces https://github.com/UN-OCHA/rwint9-site/pull/439 and adds a simple custom module that store account changes in a table and use it to show the history on a user edit form for people with sufficient rights (i.e. editors).

**Note:** Also included for the future but not used in this PR, is the addition of the hooks for the reliefweb_revisions module from https://github.com/UN-OCHA/rwint9-site/pull/439 because this is a valuable improvement IMO.

### Tests

1. Checkout this branch and run `drush cim` to enable the module.
2. Log in with a user with the "User Manager" role (A)
3. Log in in another tab (for convenience) with a normal user (B)
4. As (B), edit your account (ex: change display name)
5. As (A), open (B)'s edit page, check that the history appears at the bottom with an initial record ("Account creation" message) with and another more recent record with the display name change.
6. Make so change to (B)'s account as (A) and check those appears also in the history on (B)'s edit form.

Ex:
Screen Shot 2023-01-12 at 15 56 30


left23 (approved):
I don't quite get the expected results. I can see the User History when I am an Admin user, or when I set the User A and B to also have Editor role.

I am unclear if:
1. User A should have  _only_  the User Manager role, or are other roles needed too?  
2.  A normal user [B] is one that is active but without extra roles?

When I set User A and B to include the Editor role I can see the User History and I can confirm it works as described in the PR description.

I'm approving this so as not to block anything as this might be an issue with my understanding of what roles are needed.

Please let me know if I should look at this again.

orakili:
@left23 Sorry for the invalid instructions. They should have read:

(A) should have the "Editor" role to be able to see the "Revisions" block at the bottom of a user edit form and the "User manager" role to be able to access the edit form of other users.

(B) shouldn't have any role (but it doesn't matter if it has some). 

So you were correct to add the "Editor" role to (A). Merging as you get the correct behavior with it.

PR #544, opened 2023-01-16
RW-731 Break anchors to prevent overflow

left23:
Refs: RW-731

### DIFF
--- a/html/themes/custom/common_design_subtheme/components/rw-document/rw-document.css
+++ b/html/themes/custom/common_design_subtheme/components/rw-document/rw-document.css
@@ -96,6 +96,11 @@
   }
 }
 
+/* Ensure long URLs in WYSIWYG wrap */
+.rw-article__content a {
+  word-break: break-all;
+}
+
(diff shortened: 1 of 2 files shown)


orakili (dismissed):
Approving because that fixes the issue from the ticket but I think this rule and the title one above should be moved to https://github.com/UN-OCHA/rwint9-site/blob/develop/html/themes/custom/common_design_subtheme/components/rw-document/rw-document.css which is the component for the "document" article pages (report, job, training, blog post etc.) where the `how to apply` and `fee information` sections appear while the `rw-opportunities` is the component for the "Opportunities" block on the homepage.

Also I wonder if the `break-all` should not be applied to all links inside the `.rw-article__content` class which wraps the text of those "document" article pages because some times job posters put the same link in the body and in the how to apply section. What do you think?

left23:
@orakili Yes, thanks for pointing that out. I've made some changes.

orakili (approved):
Thanks!

PR #545, opened 2023-01-17
[RW-732] Fix issue with image field

orakili:
Refs: RW-732

https://www.drupal.org/node/3293812 introduced a change to the Drupal core ajax.js file to try to prevent a script from running before its dependencies are loaded.

This does not work well when such a dependency has the `nomodule` attribute as it seems to prevent the execution of the rest of the ajax command (ex: 1. add dependencies, 2. replace widget (not executed if (1) contains `nomodule`).

This affects the image fields on RW because the widget loads the `cd-table.js` script which depends on the `array-from-ponyfill.js` that has the `nomodule` attribute, resulting the "create image" button to fail to replace the field widget.

This PR removes the `array-from-dependency.js` (as it was for IE 11 compatibility). It also fixes the ajax override for disable the submit buttons as now, it is expected to return a promise.

### Tests

1. Checkout the latest develop or v1.10.0
2. Run `composer install` to update to Drupal 9.5.1
3. Run `drush cr` and `drush cim`
4. Go to `/node/add/report` then the `Headline` section
5. Click the "Add image" button, upload an image, fill in the copyright field
6. Click the "Create image", there may be a spinner for a short time but the widget should not change (<-- the bug)

**Aftter checking out this branch**

1. Run `drush cr`
2. Go to `/node/add/report` then the `Headline` section
3. Click the "Add image" button, upload an image, fill in the copyright field
4. This time clicking "create image" should properly update the widfget


orakili:
@left23 Sorry, could you kindly try to review this today so I can include it in the deployment tomorrow?

left23 (approved):
Works as described.

PR #549, opened 2023-01-20
[RW-736] Add training language backend filter and restrict autocomplete suggestions

orakili:
Refs: RW-736

This adds a "Course/Event language" filter to the training moderation backend form, renames "Language" to "Advertisement Language" and limits the languages and themes returned by the autocomplete, to be consistent with the training form.

Excluding irrelevant values from the autocomplete suggestions is also applied to the theme and country filters for jobs to be consistent with the job form.

### Tests.

*Note:  make sure you have a recent enough DB so the `state` has the correct values*

1. Check out this branch
2. Clear the cache
3. Open `/moderation/content/training`
4. Check that the "Advertisement Language" and "Course/Event Language" filters are present in the dropdown
5. Select "Advertisement Language", type "Other", there should be no suggestion, type "Spanish", it should appear
6. Select "Course/Event Language", type "Other", it should appear, type "Spanish", it should also appear

### DIFF
--- a/html/modules/custom/reliefweb_moderation/src/Services/TrainingModeration.php
+++ b/html/modules/custom/reliefweb_moderation/src/Services/TrainingModeration.php
@@ -7,6 +7,7 @@ use Drupal\Core\Session\AccountInterface;
 use Drupal\reliefweb_moderation\EntityModeratedInterface;
 use Drupal\reliefweb_moderation\Helpers\UserPostingRightsHelper;
 use Drupal\reliefweb_moderation\ModerationServiceBase;
+use Drupal\reliefweb_utility\Helpers\ReliefWebStateHelper;
 use Drupal\reliefweb_utility\Helpers\UserHelper;
 
 /**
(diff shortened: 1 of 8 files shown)


html/modules/custom/reliefweb_moderation/src/Services/TrainingModeration.php
```diff
@@ -302,6 +303,15 @@ protected function initFilterDefinitions(array $filters = []) {
     $definitions['career_categories']['label'] = $this->t('Professional function');
     $definitions['training_type']['label'] = $this->t('Category');
     $definitions['training_format']['label'] = $this->t('Format');
+    $definitions['theme']['exclude'] = ReliefWebStateHelper::getTrainingIrrelevantThemes();
+    $definitions['language']['label'] = $this->t('Advertisment language');
```
emmajane, line 307: Oops. Missing an "e" on "Advertisement"

PR #552, opened 2023-01-27
[RW-739] Allow changing attachment download file name.

orakili:
Refs: RW-739

This updates the report attachment file widget to allow changing the download file name (the filename part of the download URL, the downloaded file stays the same).

To do so, a "view file name details" is added below the file link in the widget to display the uploaded file name (to ease identify what file was last uploaded for the field item) and a field to change the download file name:

Screen Shot 2023-01-27 at 18 43 00

The filename is both validated in the frontend and in the backend.

### Tests

1. Checkout this branch
2. Clear the cache
3. Log in as an Editor
4. Edit or create a report, attach some files and save as published and take notes of the file URLs in the report page
5. Edit the report, click "view file name details" for an attachment, change the download file name and save
6. Check that the `filename` part of the URL for this attachment changed to the new file name.
7. Check the revisions at the bottom of the page to confirm the file name change.

To go further, edit the report, attach a new file, change the file name (do not save yet), replace a file, re-order them etc. and after saving ensure that the attachments on the report page match the attachments in the edit form (order, download file names etc.).


left23 (approved):
Works as described. I had issues with the placement of the error message in Firefox on Ubuntu (the error displays at the very top of my browser window, regardless of where I am on the form), so at first it was hard to understand that the `Download file name` field was presenting an error and preventing the page save. Unless @lazysoundsystem reports something similar, I don't think it's a blocking concern.

lazysoundsystem:
Works for me too, but I could also reproduce Lisa's unusually-placed error message - I don't think I've ever seen one like it - when I input a filename that didn't include an extension.
In Firefox on Ubuntu, it comes at the top of my screen in the middle - over the address bar when the browser window is full size but outside the window completely when it is placed further down on the screen.

I tried in Brave too (again on Ubuntu) and the behavior is a bit better: it jumps to the element and shows an error message below for a few seconds.

I tried with intentional errors on other fields but they react with 'the usual' error messages.

orakili:
@left23 @lazysoundsystem Could you attach a screenshot because I'm not sure which error message you are referring to?

left23:
@orakili It's captured in the screenshot in the description. The Download File Name field has a red border, and a tooltip error message. The tooltip is what we are referring to.

lazysoundsystem:
Well, I went back to get a screenshot and now I see the tooltip, in Firefox, against the field with the error, and the form scrolls appropriately:
image

Yesterday, the same tooltip escaped the browser window altogether and appeared at the top of my screen.

But. Then, I tried to take a screenshot of an error message when I put in an invalid date. As I hadn't corrected the invalid file name, I managed to trigger the original 'out-of-browser' tooltip, which again, it took me some time to see:
image

When I fixed the invalid filename and again tried to submit with an invalid date, I got the 'usual' error message for that field I was expecting:
image

lazysoundsystem:
Just seen your comment, @orakili - I agree it's a minor thing and that Firefox users are unlikely to hit it, but it is (probably acceptably so) a bit inconsistent with the rest of the form.

I went back for another screenshot to say that the tooltip also tells me - while I'm filling in the form - whether the value is valid or not, but it took a while to trigger that message and when I did get it, it was inaccurate:
image

left23:
@orakili I agree. If it's an impediment, a Content Editor will log it. Meanwhile FF might improve the implementation.

left23:
@orakili I'm happy to test more too, based on what @lazysoundsystem discovered. Let me know if so.

orakili:
@left23, @lazysoundsystem Thanks. I don't think it's worthy spending more time on this. Editors will create a new ticket if they are bothered by the validation message.

PR #557, opened 2023-02-03
[RW-741] allow unselecting the advanced mode checkbox in rivers

orakili:
Refs: RW-741

This fixes the logic regarding the advanced search checkbox which was preventing it from being deselected.

### Tests

1. Check out this branch
2. Go to a river page
3. Check the "advanced search" checkbox at the bottom of the filters
4. Uncheck it
5. Add a filter (ex:  country)
6. Check the "advanced search" checkbox
7. Uncheck it -> there should be a popup asking for confirmation. Press cancel and confirm nothing changed and the checkbox is still checked.
8. Repeat (7) but this time press "ok", this should reload the page with the filter gone and "advanced search" should be unchecked


left23 (approved):
Works as described

PR #558, opened 2023-02-03
[RW-740] Extended text trimming

orakili:
Refs: RW-740

This updates the text helper to trim all Z and C unicode categories (notably to remove \u200b (zero width space) and trim the taxonomy term names when building the letter navigation on the /organizations page.

### Tests

1. Check if you have the "CorpsAfrica" source (/organization/corpsafrica), if not grab a recent light DB dump
2. Clear the cache
3. Check the `/organizations` page, there should be a blank space in the letter navigation between C and D

Screen Shot 2023-02-03 at 11 31 44

1. Clear the cache
2. Go to `/organizations`, the blank "letter" should be gone
3. Edit the `/organization/corpsafrica` source, copy the name in your text editor (or something that can show invisible characters like your terminal), it should show that the name starts with `\u200b` (ex: `<200b>CorpsAfrica`).
5. Do not change anything in the form and save it as "active".
6. Repeat (3) but this time, the name should be simply `​CorpsAfrica` without the starting `\u200b`.


lazysoundsystem (approved):
Works as described.

PR #560, opened 2023-02-05
fix: Rename the `beta` role to the expected `beta_tester` so access checks work.

cafuego:
And ensure that all users with the role assigned are updated, too.

Refs: RW-743


attiks (approved):
Looking good

lazysoundsystem (dismissed):
Tested this locally and can see draft topics with a user with the 'beta-tester' role.

(Note for other testers: I had to uninstall the 'reliefweb_user_history' module before checking out this branch so that I could import the config settings).

orakili (approved):

PR #567, opened 2023-02-10
[RW-700] Better distinction between rivers, sub rivers etc.

orakili:
Refs: RW-700

This adds unique page title and canonical URLs to rivers with 1/2 filters or view. See ticket for longer explanation of the approach.

### Tests

1. Checkout the branch and clear the cache
2. Go to a country page, click on a view all xxx maps and infographics for example, check the page title and canonical URL which could contain a `list` parameter with the page title
3. Click on a page with the pager, check that the canonical URL is the same + the `page` parameter
4. Add a couple more filter or a search query and check that the page title is set to `Updates` with `https://reliefweb.int/updates` as canonical URL


left23 (approved):
I _think_ I am getting the intended results. I noticed the Advanced Search params are still as a code in the URL (Eg. `updates?view=maps&advanced-search=(PC13)&search=health`) but this seems to be the intention (based on the JIRA comments). The other params as described in the PR description work as described.

orakili:
@left23 Thanks. Keeping the advanced-search is indeed the intention.

PR #568, opened 2023-02-14
[RW-746] Use event date for sorting disasters

orakili:
Refs: RW-746

This uses the `event` date to sort disasters to match the changes in the API from RW-594.

The disasters are now sorted by event date first and then ID (to have most recently created one first) for those with the same event date.

### Tests

1. Update the most recent disaster (latest ID), set the event date to yesterday and save
2. Update the second most disaster (second lastest ID), set the event date to today and save
3. Confirm that the homepage and /disasters page have the same order with (2) first and (1) second
4. Update (1), set the event date to today and save
5. Confirm that the homepage and /disasters page have the same order with (1) first and (2) second


lazysoundsystem (approved):
Works as described.

PR #570, opened 2023-02-14
[RW-747] pecify revision creation time when manually creating revisions

orakili:
Refs: RW-747

This explicitly sets the revision creation time when creating a revision manually. Without that the previous revision timestamp is used.

### Tests

1. Edit a job source (allowed content format = job), save it as inactive, there should be revision entry with "inactive"
2. Create a job with that source, publish it
3. Check the source history, there should be a new revision entry with the "active" status, the date for this revision should be the same as the inactive one

Repeat the above but this time, the latest "active" revision should have a different timestamp than the "inactive" revision before it.


left23 (approved):
Works as described

PR #571, opened 2023-02-15
[RW-748] Add github action to run tests

orakili:
Refs: RW-748


orakili:
Finally, no choice anymore, we to deal with the 10 minutes long tests as travis has been disconnected.

cafuego (approved):
Looks like it should do the trick, even if slowly.

PR #577, opened 2023-03-07
[RW-751] Allow original extension in download file name validation pattern

orakili:
Refs: RW-751

This fixes an issue with the download file name of report attachments when the extension is not lower case.

### Tests

1. Checkout the branch and clear the cache
2. Edit or create a report and fill in the mandatory fields
3. Upload an attachment with the extension in **upper case** (ex: .PDF)
4. Check that the download file name (under "view file name details") is fine with either `PDF` or `pdf` as extension.
5. Leave it as `PDF` (upper case) and save the report and download the file to confirm it has the `PDF` extension
6. Edit the report again, change the extension to `pdf` (lower case) and save and download and check the extension is `pdf`
7. Edit the report again, this time, check that `PDF` (upper case) is not allowed (expected behavior because the download file name's original extension is now `pdf`).

### DIFF
--- a/html/modules/custom/reliefweb_files/src/Plugin/Field/FieldType/ReliefWebFile.php
+++ b/html/modules/custom/reliefweb_files/src/Plugin/Field/FieldType/ReliefWebFile.php
@@ -1280,19 +1280,40 @@ class ReliefWebFile extends FieldItemBase {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


left23 (approved):
Works as described. I get an `invalid characters or extension` error after I change the lowercase file extension to uppercase.

PR #582, opened 2023-03-17
[RW-753] Use proper value for filter for source publications API queries

orakili:
Refs: RW-753

This fixes the filter to get the list of publications for the sources.

Screenshot 2023-03-17 at 11 36 10

## Tests

1. Checkout this branch and clear the cache
2. Go to /organizations, and check that some organization show something like the screenshot above (may need to use letter filter or pagination to find a source with content when using the slim DB)


left23 (approved):
Works as described.

PR #584, opened 2023-03-17
[RW-755] Fix access to taxonomy term pages

orakili:
Refs: RW-755

This fixes the access to taxonomy term pages (all but countries, disasters and sources) that are supposed to be only accessible to people who can edit them.

## Tests

1. Checkout the branch
2. Try to access term pages (ex: /career-categories/administrationfinance, /country/afg) as an anonymous user --> should get a 404 for all pages except for countries, disasters and sources.
3. Do the same as a normal authenticated user -> same results
4. Do the same as a user with the Editor role -> same results
5. Do the same as a user with the Webmaster role -> should be able to access the term pages


left23 (approved):
Works as described.

PR #589, opened 2023-03-29
chore: add and configure environment indicator

lazysoundsystem:
Refs: OPS-8939

Should be consistent with the changes to all the other repos. No permissions included to see the environment switchers and requires ansible changes to see the indicators on the dev and prod sites.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "drupal/core-composer-scaffold": "^9.3",
         "drupal/core-dev": "^9.3",
         "drupal/core-recommended": "^9.3",
+        "drupal/environment_indicator": "^4.0",
         "drupal/geofield": "^1.39",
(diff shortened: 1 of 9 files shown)


orakili (commented):
@lazysoundsystem The `environment_indicator` module is only enabled on dev/local via the config_split settings. So I'm wondering a few things:

1. Is the `config/environment_indicator.switcher.prod.yml` needed?
2. Shouldn't the `config/environment_indicator.switcher.*.yml` be moved to `config_dev`?
3. Is it ok to expose the dev URL in `config/environment_indicator.switcher.dev.yml`? Shouldn't this be set via ansible for example?
4. What about other instances like the `demo` one?

lazysoundsystem:
Closing for https://github.com/UN-OCHA/rwint9-site/pull/636

PR #613, opened 2023-06-26
[RW-772] Date filters

orakili:
Refs: RW-772

This adds and renames date filters on the country, disaster and source moderation backends.

### Tests.

2. Go to `/moderation/content/country`, check that there is a "Change date" filter. Select a date, filter and confirm that it matches the "updated" column
3. Go to `/moderation/content/disaster`, check that there is a "Disaster date" "change date" filter. Select a date, filter and confirm that it matches the "disaster date" column
4. Go to `/moderation/content/disaster`, check that there is a "Creation date" "change date" filter. Select a date, filter and open a disaster in the results, check the revision history at the bottom and confirm the disaster was indeed created on the selected date
5. Go to `/moderation/content/source`, check that there is a "Creation date" "change date" filter. Select a date, filter and open a source in the results, check the revision history at the bottom and confirm the source was indeed created on the selected date

### DIFF
--- a/html/modules/custom/reliefweb_moderation/src/ModerationServiceBase.php
+++ b/html/modules/custom/reliefweb_moderation/src/ModerationServiceBase.php
@@ -845,6 +845,14 @@ abstract class ModerationServiceBase implements ModerationServiceInterface {
           'form' => 'omnibox',
           'widget' => 'datepicker',
         ],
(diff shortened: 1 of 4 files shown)


lazysoundsystem (approved):
All works as described.

PR #632, opened 2023-07-30
[RW-734] Script to remove the HR.info origin URLs

orakili:
Refs: RW-734


orakili:
Closed in favor of https://github.com/UN-OCHA/rwint9-site/pull/633.

PR #635, opened 2023-07-31
[RW-785] Local stack

orakili:
Refs: RW-785

This is against the branch for the D10 upgrade of RW which should become mainstream soon.

Local stack inspired by and improving on the https://github.com/UN-OCHA/drupal-starterkit one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,10 @@ buildlog.txt
 /private/
 /notes/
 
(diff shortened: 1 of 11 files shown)


lazysoundsystem (approved):
Tested locally and all works okay.

Will put a .env.example file in the starterkit to make the set-up more consistent.

PR #636, opened 2023-07-31
chore: re-add env indicator config

lazysoundsystem:
Refs: OPS-8939

A belated follow-up to https://github.com/UN-OCHA/rwint9-site/pull/589 - on a new branch due to merge conflicts.

The questions on that PR:

* _Is the config/environment_indicator.switcher.prod.yml needed?_
Not really, it's being added as standard, as some will be using it, but by default it's not enabled.
* _Shouldn't the config/environment_indicator.switcher.*.yml be moved to config_dev?_
Putting the local one in config_dev - I don't think config_dev is used for the dev and demo servers. 
* _Is it ok to expose the dev URL in config/environment_indicator.switcher.dev.yml? Shouldn't this be set via ansible for example?_
Thanks for pointing this out, I'm removing all the urls but for local, moving them to ansible. 
* _What about other instances like the demo one?_
Now included.


orakili:
@lazysoundsystem I'll review and update composer after https://github.com/UN-OCHA/rwint9-site/pull/563 is merged to avoid having to the work twice if that's ok.

lazysoundsystem:
Yes, no rush on this at all. We've survived without it for the last 25 years...

orakili (approved):
Finally, took the time to test the PR. It seems to work well. Thanks.

PR #645, opened 2023-08-31
[RW-803] Webp and responsive images

orakili:
Refs: RW-803

This adds `webp` conversion pipeline for images and set up basic responsive image styles.

Basically this uses **existing** smaller image styles on  small screens for report images, map/infographic attachment previews etc. in addition to offering `webp` versions for the images.

This is for performance improvements so there are no `x2` for example (out of scope of this PR). The existing image styles were not changed either and there are no new ones to keep things simple.

We could refine further, for example there is only 1 image style (the existing one) for the announcement on the homepage which is a bit large on desktop and mobile but that's for another ticket if the changes in that PR are not enough.

## Tests.

**Prerequisites** 

Ensure you have `stage_file_proxy` properly set up for your local instance. If not, an option is to create a new site instance with `./local/install.sh` (see the `local/README.md` file for instructions) and importing a light DB dump. You would need also an instance of the [RW API](https://github.com/UN-OCHA/rwint-api) (same install process).

Otherwise create a few new reports (map with attachment, situation report with attachment, report with an image, headline with image and set it on the homepage) to check the display of their attachments and image on their page.

Note: with `stage_file_proxy` it may take a bit of time to download the original images and generate the webp ones.

**Actual tests**

1. Checkout the branch, run `composer install`, clear the cache, import the config.
2. Visit the homepage, confirm the headline and announcement image appears. Check the page source to see if the responsive image markup is present and check the network tab of the dev tools to see what image the browser uses.
3. Repeat with the `/updates` page
4. Repeat with the `/country/afg` page for example


left23:
I'm having too many set up problems to get the actual PR part. I'm trying to get the local site set up working for both the API and rwint9 - I have rwint9-site running with the lite db imported and config imported. I see content in the backend but no homepage or list content.
The RW API site is running locally, I can visit the domain, but I get a time out when I run `drush rapi-i report` which I understand to should generate the content. As far as I can see, the `settings.site.php` has the correct API test site.
Giving up for now.

left23 (approved):
I'm approving this without testing. I'm happy to review it properly once I get the set up correct but don't want to block this, especially before next week.

left23:
@orakili Thanks for the clarification around the local site set up. I've better tested this PR now.

From what I can see, the /updates page doesn't use Responsive images but I understand that to be intentional.
I've checked a country page and can see the Responsive images there.
On the homepage, the Latest headlines are displaying the Responsive images and the announcement is not which I understand to be working as intended.

PR #654, opened 2023-09-27
[RW-812] Fix external links for jobs/training

orakili:
Refs: RW-812

This adds the target and rel attributes to external links to open in new tab for links inside fields using the markdown_editor (ckeditor 5) text format.

## Tests.

1. Checkout the branch, clear the cache, import the config
2. Create or edit a job or training, add some links in the body starting with `https://reliefweb.int` and your local domain like `https://rwint-local.test`.
3. Add links to external sites
4. Save
5. Confirm that the RW (prod and local) links do not have a target and rel attributes
6. Confirm that the external links have those attributes and open in new tab/window

### DIFF
--- a/config/filter.format.markdown_editor.yml
+++ b/config/filter.format.markdown_editor.yml
@@ -3,7 +3,10 @@ langcode: en
 status: true
 dependencies:
   module:
+    - editor
+    - media
     - reliefweb_fields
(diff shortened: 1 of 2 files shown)


left23 (approved):
Works as described.

PR #662, opened 2023-10-06
[OPS-9699] Remove mapbox token from mabox responses

orakili:
Refs: OPS-9699

This substitutes the mapbox token in the proxied mapbox responses.

## Tests

1. Checkout the branch
2. Recreate the drupal container
3. Visit `/disasters`
4. Check the network tab (refresh if necessary), filter by `mapbox`
5. Check the content of any json file, the mapbox token should not be present (instead there should be `access_token=token`)

### DIFF
--- a/docker/etc/nginx/custom/03_mapbox.conf
+++ b/docker/etc/nginx/custom/03_mapbox.conf
@@ -20,13 +20,25 @@ location /mapbox/ {
   ## Remove cookies.
   more_clear_input_headers Cookie;
 
+  ## Ensure the proxy host is in a variable, so nginx will not cache the IP
+  ## indefinitely and cause 502 errors if the IPs change.
+  set $mapbox_host "api.mapbox.com";
+
   ## Proxy the request to the mapbox API.
   proxy_ssl_server_name on;
-  proxy_set_header Host "api.mapbox.com";
-  proxy_pass https://api.mapbox.com$mapbox_request_uri;
+  # Disable compression so we can substitute the token in the response body.
+  proxy_set_header Accept-Encoding "";
+  proxy_set_header Host $mapbox_host;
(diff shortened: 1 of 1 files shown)


cafuego (commented):
I've checked it on unocha.org dev and the `proxy_set_header Accept-Encoding "";` is obviously what I was missing. The replace rule then works, but the links to the tile servers do not get re-rewritten, as they're not proxied, so the library is unable to download tiles: `{"message":"Not Authorized - Invalid Token"}`

orakili:
@cafuego Where do you see this `{"message":"Not Authorized - Invalid Token"}` message?

From my tests, all the tiles load properly. 

The token substitution is done in the response of `/mapbox/v4/reliefweb.xxx.json?style=mapbox://styles/reliefweb/xxx&secure&access_token=token` file for the tile URLs:

```
https://a.tiles.mapbox.com/v4/reliefweb.xxxx/{z}/{x}/{y}.vector.pbf?access_token=token&style=mapbox://styles/reliefweb/xxx
https://b.tiles.mapbox.com/v4/reliefweb.xxxx/{z}/{x}/{y}.vector.pbf?access_token=token&style=mapbox://styles/reliefweb/xxx
```

But those URLs are not called. 

The tiles requests are like `/mapbox/v4/reliefweb.xxx/1/1/0.vector.pbf?style=mapbox://styles/reliefweb/xxx&sku=yyy&access_token=token` and are properly proxied.

Zooming or scrolling loads new tiles with the `/mapbox/v4/reliefweb.xxx/1/1/0.vector.pbf?style=mapbox://styles/reliefweb/xxx&sku=yyy&access_token=token`.

cafuego:
Huh, not called?

cafuego (approved):
Merging, since the tile server URLs that get rewritten appear to not actually get used.

PR #673, opened 2023-10-27
[RW-819] Replace twitter icon

orakili:
Refs: RW-819


left23:
I came across this while browsing github for other reasons. I reviewed this locally. All good.

left23 (approved):

PR #683, opened 2023-11-20
CD-519 Remove missing UNOCHA logo from base theme via patch

left23:
chore: create patch that removes logo from common_design base theme, add to patches file

See CD-519 for screenshots and details.

Once this patch is applied, the declaration in the common_design base theme rule that sets the UNOCHA logo as a background `  background:
    linear-gradient(transparent, transparent),
    url(../../../img/logos/ocha-logo-blue.svg) center no-repeat;`  is no longer present, which means we no longer have a reference to a missing image.

This will be tested after its deployed. If it has a positive impact on SEO, we will implement this properly in the base theme as described on CD-519


orakili (approved):
Patch applies cleanly. Thanks.

PR #689, opened 2023-12-07
[RW-837] Date filter fix

orakili:
Refs: RW-837

This allows adding date filter in river without showing the datepicker and warns when the date is invalid.

Screenshot 2023-12-07 at 15 51 59

### Tests

2. Go to  `/updates` for example
3. Click on `Original publication date`
4. Enter a date in the `From` input, check that there is some indication that the date is not valid when it's not in the `YYYY`, `YYYY/MM` or `YYYY/MM/DD` format.
5. Check that the date is added only if it's valid after clicking the `Add` button.
6. Repeat (4) and (5) with the datepicker open


left23 (approved):
Works as described.

PR #694, opened 2023-12-21
chore: update drush, remove outdated sfp patch, run periodic updates

lazysoundsystem:
Refs: updates

Updated patch for https://www.drupal.org/project/drupal/issues/3008292, removed stage_file_proxy webp patch https://www.drupal.org/node/3171559 (which has gone back and forth - the present version is without `+ $paths = array_reverse($paths);` - line 14 of the removed patch - so should check if we want to include that again.)

Also needed to update drush.


lazysoundsystem:
Periodic updates will be broken until this is merged.

orakili (changes requested):
The file validation update in 10.2 requires changes to the `reliefweb_files` module: https://humanitarian.atlassian.net/browse/RW-843.

orakili:
Replaced by https://github.com/UN-OCHA/rwint9-site/pull/703.

PR #698, opened 2024-01-10
Rw 821

attiks:


attiks:
@orakili this is failing on install, but no idea why, I thought I saw it mentioned somewhere but couldn't find it

orakili (approved):
Working well. 

I added a commit to fix a "Warning: Attempt to read property "entity" on null" for entities without a country like remote jobs or online training.

PR #708, opened 2024-01-15
RW-841 Update base theme to v9.3

left23:
This specifically avails of a recent update to the base theme to load fonts in a non-blocking manner. This was included in base theme release v9.2.3
See https://github.com/UN-OCHA/common_design/pull/448 for Font testing steps

There has since been another update for v9.3 which uses the SDC and rearranges the directory structure in the base theme.
https://github.com/UN-OCHA/common_design/releases/tag/v9.3.0

We shouldn't need to do anything to react to the v9.3 update since we do not reference any component's location using the directories, but instead use library extends when interacting with base theme components.

Refs: RW-841

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         "reliefweb/simple-autocomplete": "^v1.3",
         "reliefweb/simple-datepicker": "^v1.3",
         "symfony/uid": "^6.2",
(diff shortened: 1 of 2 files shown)


orakili (approved):
Thanks.

PR #713, opened 2024-01-18
CD-519 Take 2 - Remove missing UNOCHA logo from base theme via patch

left23:
chore: base theme patch to remove ocha logo bg img

Refs: CD-519

See https://github.com/UN-OCHA/rwint9-site/pull/683


orakili (approved):
Thanks.

PR #728, opened 2024-02-20
feat: Add an override to allow PUT on the new API handler in Drupal.

cafuego:
Refs: RW-892


docker/Dockerfile
```diff
@@ -62,6 +62,7 @@ COPY --from=builder /srv/www/composer.patches.json /srv/www/composer.patches.jso
 COPY --from=builder /srv/www/composer.lock /srv/www/composer.lock
 COPY --from=builder /srv/www/PATCHES /srv/www/PATCHES
 COPY --from=builder /srv/www/scripts /srv/www/scripts
+COPY --from=builder /src/www/docker/etc/nginx/map_block_http_methods.conf /etc/nginx/map_block_http_methods.conf
```
cafuego, line 65:
```suggestion
COPY --from=builder /srv/www/docker/etc/nginx/map_block_http_methods.conf /etc/nginx/map_block_http_methods.conf
```

PR #734, opened 2024-02-27
Job tagger

attiks:
For testing


lazysoundsystem:
Noticed the db version bump - just fyi, we're now using mysql 10.11 for tests and in the local stacks - https://humanitarian.atlassian.net/browse/OPS-10092

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -0,0 +1,115 @@
...
+        'select[name="field_source"]' => $conditions,
+      ],
+    ];
+  }
+}
+
+/**
+ * Implements hook_entity_base_field_info().
+ */
+function reliefweb_job_tagger_entity_base_field_info(EntityTypeInterface $entity_type) {
```
orakili, line 100: @attiks Wouldn't be better to use a bundle field instead of a base field so that it's only added to the jobs and not to all the nodes?

attiks: True, fixed

orakili (changes requested):
Some issues I encountered while testing (in random order):

**Jobs submitted before module is enabled**

When editing an already submitted job as a non trusted, non editor user, it's not possible anymore to edit the previously selected career, theme, experience fields but at the same time the job is not queued.

That means it's not possible for job posters to edit those fields in their pending submissions after the module is enabled.

Scenario:

- User A submits a job on March 12 with some theme, carreer and experience. The theme and career category are wrong  (ex: misclicked).
- The module is enabled on March 13. 
- User A notices that the theme is wrong and decide to edit its submission since it's still pending --> the fields are not visible and editable anymore.

**Delay between submission and AI tagging**

The actual tagging being done on cron, there is a delay between the time of submission and the time the job is actually ready for review by the jobs team. This can probably be "solved" via guidelines but some message in the revision log saying the document is queued for tagging may help as well.

**Cron**

Processing the queued jobs on cron is quite slow, potentially delaying other things like the email notifications. Maybe it would be better to have a separate drush command that we can run at other intervals.

**PHP warnings in the job tagger**

```
 [info] Only variables should be passed by reference OchaAiJobTagTaggerWorker.php:98 
 [info] Only variables should be passed by reference OchaAiJobTagTaggerWorker.php:111
 [info] Only variables should be passed by reference OchaAiJobTagTaggerWorker.php:124
```

**Logging**

There are not many log entries about what is going on with the job tagging. For example, it's not mentioned when a job is queued for tagging.

**Disclaimer**

The `The process of posting jobs via ReliefWeb is now even smoother. The form now includes an automated function for the "career", "job experience" and "themes" categories; fewer fields for users to fill and higher efficiency overall.` disclaimer message tends to disappear when editing a document.

It appears for a brief instant then disappears.

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -0,0 +1,164 @@
...
+      '#text' => t('The process of posting jobs via ReliefWeb is now even smoother. The form now includes an automated function for the "career", "job experience" and "themes" categories; fewer fields for users to fill and higher efficiency overall.'),
+      '#format' => 'html',
+      '#weight' => -90,
+      '#prefix' => '<div class="messages messages--warning cd-alert cd-alert--warning">',
+      '#suffix' => '</div>',
+      '#states' => [],
+    ],
+  ];
+
+  $sources = array_keys(UserPostingRightsHelper::getUserPostingRights());
```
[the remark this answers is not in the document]
attiks: Fixed

attiks:
To be tested with local install of https://github.com/UN-OCHA/ocha_ai develop branch

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -0,0 +1,261 @@
...
+    case 'node_insert':
+      $group = $implementations['reliefweb_job_tagger'];
+      unset($implementations['reliefweb_job_tagger']);
+      $implementations['reliefweb_job_tagger'] = $group;
+      break;
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
```
orakili, line 162:
```suggestion
 * Implements hook_ENTITY_TYPE_presave().
```

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -0,0 +1,261 @@
...
+    return;
+  }
+
+  if ($node->hasField('reliefweb_job_tagger_status') && $node->get('reliefweb_job_tagger_status')->value == 'queue') {
+    reliefweb_job_tagger_queue_job($node);
+
+    $node->set('reliefweb_job_tagger_status', 'queued');
+    $node->setRevisionLogMessage('Job has been queued for tagging');
+    $node->setNewRevision(FALSE);
+    $node->updateLoadedRevisionId();
```
orakili, line 226: The [comment for this method](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21RevisionableInterface.php/function/RevisionableInterface%3A%3AupdateLoadedRevisionId/10) says: "This method should not be used, it could unintentionally cause the original revision ID property value to be lost." Is it necessary? If yes, could add a comment in the code explaining why?

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -0,0 +1,261 @@
...
+    $node->save();
```
orakili, line 227: Normally an entity should not be saved again in its own insert/update/delete hooks because it's still inside the DB transaction for the orignal `save()` call.

RW introduces an `after_save` hook in `Drupal\reliefweb_entities\BundleNodeStorage` that can be used for that. See `reliefweb_api_entity_after_save()` for an implementation example.

Otherwise, probably better to move the logic to the `reliefweb_job_tagger_node_presave` and remove the `$node->save()` altogether.

Same comment for the `reliefweb_job_tagger_node_update` below.

html/modules/custom/reliefweb_job_tagger/src/Plugin/QueueWorker/OchaAiJobTagTaggerWorker.php
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    $nid = $data->nid;
+
+    if (empty($nid)) {
+      \Drupal::logger('deubg')->notice('no nid');
```
orakili, line 65:
```suggestion
      \Drupal::logger('debug')->notice('no nid');
```

Other instances of this typo below.

html/modules/custom/reliefweb_job_tagger/src/Plugin/QueueWorker/OchaAiJobTagTaggerWorker.php
```diff
@@ -0,0 +1,255 @@
...
+    if ($needs_save) {
+      if ($node->hasField('reliefweb_job_tagger_info')) {
+        $node->set('reliefweb_job_tagger_info', [
+          'value' => implode("\n\n", $message),
+          'format' => 'markdown',
+        ]);
+      }
+      $node->revision_log = 'Job has been updated by AI.';
+      $node->set('reliefweb_job_tagger_status', 'processed');
+      $node->isNewRevision();
```
orakili, line 166:
```suggestion
      $node->setNewRevision(TRUE);
```

I think this is meant to be a call to `setNewRevision`.

orakili (changes requested):
Seems to be working well enough but not approving yet because there a few things to change (see individual comments).

config/field.field.node.job.field_job_experience.yml
```diff
@@ -12,7 +12,7 @@ entity_type: node
 bundle: job
 label: 'Job years of experience'
 description: ''
-required: true
+required: false
```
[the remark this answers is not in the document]
attiks: Fixed using css, I might look for a better way to handle it.

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
attiks: It's needed, without it an exception is thrown.
Comment added.

html/modules/custom/reliefweb_job_tagger/src/Plugin/QueueWorker/OchaAiJobTagTaggerWorker.php
attiks: Removed

html/modules/custom/reliefweb_job_tagger/src/Plugin/QueueWorker/OchaAiJobTagTaggerWorker.php
attiks: True, fixed

html/modules/custom/reliefweb_job_tagger/src/Plugin/QueueWorker/OchaAiJobTagTaggerWorker.php
```diff
@@ -0,0 +1,255 @@
...
+        continue;
+      }
+
+      if ($score > .25) {
+        $result[] = $theme;
+      }
+
+    }
+
+    return $result;
```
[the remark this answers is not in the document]
attiks: Re-factored

config/field.field.node.job.field_job_experience.yml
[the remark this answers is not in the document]
attiks: In the mean time, years of experienced has been removed

html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
attiks: Switched to reliefweb_api_entity_after_save

orakili (commented):
See https://github.com/UN-OCHA/rwint9-site/pull/768 that contains adjustments.

attiks:
Now depends on https://github.com/UN-OCHA/ocha_ai/pull/25 for cache tags

orakili (approved):

PR #749, opened 2024-03-15
feat: Remove crisis figures

attiks:
Refs: #RW-896


rupl:
There are some orphaned theme functions and templates associated with this code removal. I know specifically of `html/modules/custom/reliefweb_key_figures/templates/reliefweb-key-figures.html.twig`

attiks:
Next deploy everything will be removed

rupl (approved):
Got it, in that case have a green check mark!

PR #750, opened 2024-03-19
[RW-921] Update ocha_ai module

orakili:
Refs: RW-921

Latest version of the `ocha_ai` module with styling fixes for the chat.


rupl (approved):
Works like normal 👍🏼

PR #752, opened 2024-03-19
[RW-919] Better wrapping of links

orakili:
Refs: RW-919

Change the css rule to wrap text in links to use the more recent `overflow-wrap: anywhere` that tries not to break words if possible.

See ticket for screenshots of what to expect.

## Tests

2. Edit/create a node, add some text with (a) a very long link containing words and another paragraph (b) with a very long URL.
3. Save or preview and confirm that for (a) the words are not broken and that the link text wraps at separators like spaces and that (b) is broken at some places in the URL and doesn't overflow.


rupl (approved):
LGTM

Screenshot 2024-03-19 at 09 22 47

PR #757, opened 2024-03-21
test: See if the newer docker images still have the xdebug problem.

cafuego:
Refs: OPS-10095


cafuego:
Apparently they do not.

PR #759, opened 2024-03-25
chore: use gtm_barebones

lazysoundsystem:
Refs: RW-856

This leaves google_tag module enabled - I did try uninstalling it at the same time, but it broke my local build (even after repeated cache-clears).

`drush pm:uninstall google_tag` executes fine, but the next page load errors with:
```
The website encountered an unexpected error. Try again later.

Drupal\Component\Plugin\Exception\PluginNotFoundException: The "google_tag_container" entity type does not exist. in Drupal\Core\Entity\EntityTypeManager->getDefinition() (line 139 of core/lib/Drupal/Core/Entity/EntityTypeManager.php). 
```
I tried `drush ev "\Drupal::entityTypeManager()->clearCachedDefinitions();"` but that didn't help.

The switch was clean on CERF https://github.com/UN-OCHA/cerf8/pull/1249/files and on CD google_tag was uninstalled first, then gtm_barebones added in a separate commit: https://github.com/UN-OCHA/common-design-site/pull/329

I'll come back to this in a bit.


lazysoundsystem:
Noting that uninstalling `reliefweb_analytics` avoids the error. Will need to update that module first.

lazysoundsystem:
Removing the check and using the standard 'dataLayer' variable name means this is no longer blocked.

orakili (approved):
Looks ok, thanks.

PR #760, opened 2024-03-26
[RW-841] Upgrade CD to v9.4.0

rupl:
Mainly to deliver the two performance improvements: fonts loaded async, and OCHA Services moved permanently to the footer for improved CLS scores.

Refs: RW-841


rupl:
@orakili I had to update the icon sprite to get the external link icon in the new OCHA Services in the footer. A few icons were added, but none removed, so I believe it was a pretty safe upgrade. If there's a part of the site you know requires extra attention I'll be happy to do more QA.

orakili (approved):
Looks fine. Thanks.

PR #768, opened 2024-04-04
[RW-931] Job tagging adjustments

orakili:
Refs: RW-931

**Note**: the ocha_ai module is set to branch from https://github.com/UN-OCHA/ocha_ai/pull/23. That will need to be changed to a tagged release after the current ocha_ai open PRs are merged.

This adjusts a few things from https://github.com/UN-OCHA/rwint9-site/pull/734.

Each commit can picked up separately in case some of the changes in that PR are not welcomed.

Some of the main changes are:

1. Removal of the `entity->save()` in the `hook_entity_after_update` because it results in unnecessary double DB writes. Setting the tagging status to "queued" in the `presave` hook and queueing the entity in the `after_save` hook is enough.
2. The use of the `instructions` field of the ai tagger config form for the message displayed on the job form. This is to be able to adjust or remove the message without a deployment. (This should be reviewed if at one point we add tagging of other content types with different "instruction" messages.)
3. The cutoff date is replaced by a check on the career category field and the tagger status field. This is more robust I think and importantly allows to test properly before the deployment.
4. There is a new form validation callback to ensure that a career category is selected when a user selects a source they are trusted for. The form `states` only affects the UI and doesn't enforce the requirement.

This was tested with several use cases:

**Existing jobs created before AI tagging introduced**

All the AI stuff is always skipped, regardless of who edits the job --> same behavior as before with career categories and themes fields visible and career categories mandatory.

**Editors**

All the AI stuff is always skipped.

**Trusted users**

Trusted source selected --> AI stuff skipped --> same behavior as before with career categories and themes fields visible and career categories mandatory --> submitting saves as published with no queueing.

Non trusted source selected --> see normal user below.

**Normal user/non trusted source**

Career categories and themes fields are not visible, message at top of form is visible.
- Saving as draft --> document not queued
- Submitting --> saves as pending and queues document for AI processing.

**New source**

Career categories and themes fields are not visible, message at top of form is visible --> saved as on-hold as expected and document is not queued for AI tagging.

### DIFF
--- a/html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
+++ b/html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
@@ -58,29 +58,33 @@ function reliefweb_job_tagger_form_node_form_alter(array &$form, FormStateInterf
     return;
   }
 
-  // Add message for job posters.
-  $form['ai_message'] = [
-    '#type' => 'fieldset',
-    '#attached' => [
-      'library' => [
-        'reliefweb_job_tagger/ai-message',
+  $settings = \Drupal::service('ocha_ai_tag.tagger')->getSettings();
+
+  if (!empty($settings['form']['instructions']['value'])) {
+    // Add message for job posters.
+    $form['ai_message'] = [
+      '#type' => 'fieldset',
+      '#attached' => [
+        'library' => [
+          'reliefweb_job_tagger/ai-message',
+        ],
       ],
-    ],
-    '#attributes' => [
-      'class' => [
-        'ai-message-wrapper',
+      '#attributes' => [
+        'class' => [
+          'ai-message-wrapper',
+        ],
       ],
-    ],
-    'message' => [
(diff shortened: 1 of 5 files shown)


html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -231,12 +271,8 @@ function reliefweb_job_tagger_entity_after_save(EntityInterface $entity) {
   }
 
   /** @var \Drupal\node\Entity\Node $entity */
-  if ($entity->hasField('reliefweb_job_tagger_status') && $entity->get('reliefweb_job_tagger_status')->value == 'queue') {
+  if ($entity->hasField('reliefweb_job_tagger_status') && $entity->get('reliefweb_job_tagger_status')->value == 'queued') {
```
attiks, line 280: If a user edits a post and save it again, the job is added to the queue again, but the worker will detect it

attiks (approved):

PR #771, opened 2024-04-09
Develop -> Main - v2.5.0

orakili:
Bumping version because of the switch to gtm barebones.

## Chores

- RW-856: Switch to GTM barebones
- RW-928: Update ocha_ai module to 1.1.1 (chat UX fixes + tagger update)
- Composer update (Drupal 10.2.5, AWS SDK 3.303.2)

## Fixes

- RW-932: fix admin menu display after regression from CD 9.4 update
- RW-928: allow both simple and detailed feedback mode for AI chat
- RW-930: better formatting of answers in AI chat


rupl (approved):

lazysoundsystem (approved):
Looks good.

And I'm going to try using the description format across all merges to main so we can keep better track of which modules get updated when.

PR #773, opened 2024-04-11
chore: Add Amazon SES support and config to be managed via Ansible.

cafuego:
Refs: OPS-9523


lazysoundsystem (dismissed):

orakili (approved):
Approving because the PR does what it means to do but marking as blocked until the code for the subscriptions is made compatible: https://humanitarian.atlassian.net/browse/RW-935.

orakili (approved):

PR #775, opened 2024-04-12
[RW-934] Fix: disable the state conditions for the OCHA production field due to incompatibility with Druapl 10.2.5

orakili:
Refs: RW-934

Temporarily always show the OCHA product field due to incompatibility of the states condition and Drupal 10.2.5 changes to the form state API which prevent posting OCHA content.


cafuego (approved):
This PR definitely disables some code.

PR #777, opened 2024-04-15
[RW-934] Fix OCHA product visibility on report form

orakili:
Refs: RW-934

This replaces the non-working Drupal form state logic with custom one to show/hide the OCHA product field when OCHA is selected.

## Tests.

1. Check out the branch, clear the cache
2. Create a new report `/node/add/report`
3. Fill in language, title, body, origin notes, country, primary country and content format
4. Check that the `OCHA product` field below the `Source` field is not present
5. Select a source other than OCHA, ex: UNDP, check that the `OCHA product` field is still not visible
6. Add `OCHA` as selection, check that the `OCHA product` field now appears
7. Remove `OCHA` from the selection, check that the `OCHA product` field disappears
8. Add `OCHA` back, remove `UNDP`, check that the `OCHA product` field is visible
9. Save
10. Edit the report, check that the `OCHA product` field is visible
11. Remove `OCHA`, add `UNDP`, check that the `OCHA product` field disappeared
12. Save
13. Edit the remove, check that the `OCHA product` field is not visible

### DIFF
--- a/html/modules/custom/reliefweb_entities/src/Services/ReportFormAlter.php
+++ b/html/modules/custom/reliefweb_entities/src/Services/ReportFormAlter.php
@@ -249,10 +249,21 @@ class ReportFormAlter extends EntityFormAlterServiceBase {
    * Ensure only 1 OCHA product is selectable and that it's mandatory when OCHA
    * is selected as source or hidden otherwise.
    *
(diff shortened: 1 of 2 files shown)


lazysoundsystem (approved):
Looks good and works as described.

PR #779, opened 2024-04-17
[OPS-9523][RW-935] Amazon SES compatibility fixes

orakili:
Refs: OPS-9523, RW-935

_This is against the [cafuego/ops-9523-amazon-ses](https://github.com/UN-OCHA/rwint9-site/tree/cafuego/ops-9523-amazon-ses) branch._

This PR adjusts the way emails are sent by the `amazon_ses` module to handle proper HTML formatting, extra headers and BCC.

**Important:** to work properly this assumes that the formatting of the message is still done by MimeMail so the `mail.profile.formatter` will have to be set to `mime_mail` for RW: https://github.com/UN-OCHA/tools-ansible/blob/977a2cc13d56c2ed66ca5f57942d660a91432201/modules/drupal/templates/settings.10.php.j2#L1276

### DIFF
--- a/html/modules/custom/reliefweb_subscriptions/src/AmazonSesMessageBuilder.php
+++ b/html/modules/custom/reliefweb_subscriptions/src/AmazonSesMessageBuilder.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Drupal\reliefweb_subscriptions;
+
+use Drupal\amazon_ses\MessageBuilder;
+
+/**
+ * Extend the Amazon SES module's message builder.
+ */
+class AmazonSesMessageBuilder extends MessageBuilder {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildMessage(array $message) {
+    // Adjust headers and expected message properties to work with the base
+    // amazon_ses message builder.
+    if (isset($message['headers'])) {
+      foreach ($message['headers'] as $key => $value) {
+        switch (strtolower($key)) {
+          case 'from':
+            $message['from'] = $value;
+            unset($message['headers'][$key]);
+            $message['headers']['From'] = $value;
+            break;
+
+          case 'reply-to':
+            $message['reply-to'] = $value;
+            unset($message['headers'][$key]);
+            $message['headers']['Reply-to'] = $value;
+            break;
+
+          case 'cc':
+            unset($message['headers'][$key]);
+            $message['headers']['Cc'] = $value;
+            break;
+
+          case 'bcc':
+            unset($message['headers'][$key]);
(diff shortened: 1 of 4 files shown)


html/modules/custom/reliefweb_subscriptions/src/ReliefwebSubscriptionsMailer.php
```diff
@@ -537,8 +542,8 @@ protected function generateEmail(array $subscription, array $data) {
       // batch. Probably not strictly necessary, but if we *do* go fast this
       // can help clear a back-log of 38,000 mails just a bit faster.
       $timer_elapsed = microtime(TRUE) - $timer_start;
-      if ($timer_elapsed < 1) {
-        usleep((1 - $timer_elapsed) * 1e+6);
+      if ($throttle && $timer_elapsed < 1) {
+        usleep((int) ((1 - $timer_elapsed) * 1e+6));
```
cafuego, line 546: Pff. What?

html/modules/custom/reliefweb_subscriptions/src/AmazonSesMessageBuilder.php
```diff
@@ -0,0 +1,81 @@
...
+
+    // Add extra headers.
+    // @see \Drupal\reliefweb_subscriptions\ReliefWebSubscriptionsMailer::generateEmail().
+    if (isset($message['params']['headers'])) {
+      /** @var Symfony\Component\Mime\Headers $headers */
+      $headers = $email->getHeaders();
+      foreach ($message['params']['headers'] as $header => $value) {
+        $headers->addHeader($header, $value);
+      }
+    }
```
cafuego, line 57: I don't understand this bit; the headers are added to the $headers class but that's not used again after that?

cafuego (approved):
Kind of a #YOLO review; I don't understand the headers bit.

However, the formatter setting is already enforced via Ansible: https://github.com/UN-OCHA/tools-ansible/blob/977a2cc13d56c2ed66ca5f57942d660a91432201/rwint-site/group_vars/production.yml#L33 (and also for development)

html/modules/custom/reliefweb_subscriptions/src/AmazonSesMessageBuilder.php
orakili: `$headers` is a reference to the Headers object of the `$email` object. So we are adding extra headers directly to the email's headers.

Those headers are then used to form the raw email when the `$email` is converted to string: https://git.drupalcode.org/project/amazon_ses/-/blob/3.0.x/src/AmazonSesHandler.php?ref_type=heads#L95

PR #783, opened 2024-04-18
[OPS-9780] Fix nginx locations for files

orakili:
Refs: OPS-9780

Same changes as https://github.com/UN-OCHA/docker-images/pull/357. See this PR's description for details.


cafuego (approved):
Nice catch.

I *think* that since we don't use multi-site and *all* properties are `default` we don't actually need the regex at all and we can just use `location /sites/default/files/`

PR #785, opened 2024-04-19
[OPS-9523] Fix missing logger channel complaint

orakili:
Refs: OPS-9523

Make Drupal happy and stop complaining about a missing service because it loads things out of order.


cafuego (approved):
Reasonable.

PR #789, opened 2024-04-24
[Rw-939] Add patch to prevent php notice from Drupal core PHP mailer

orakili:
Refs: RW-939

Tiny patch to avoid the following PHP notice when sending emails:

> str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in docroot/core/lib/Drupal/Core/Mail/Plugin/Mail/PhpMail.php line 114


attiks:
Thanks, used in AI Summarize as well

lazysoundsystem (approved):

PR #790, opened 2024-04-24
chore: local settings

lazysoundsystem:
Refs: local dev

Plus a typo and removing an obsolete line from docker-compose.yml

### DIFF
--- a/local/shared/settings/settings.local.php
+++ b/local/shared/settings/settings.local.php
@@ -177,8 +177,3 @@ if (!empty($no_cache)) {
   $settings['cache']['bins']['page'] = 'cache.backend.null';
   $settings['cache']['bins']['dynamic_page_cache'] = 'cache.backend.null';
 }
-
-$config['reliefweb_api.settings']['api_url'] = 'http://rwint-api-local-site/v1';
-$config['reliefweb_api.settings']['api_url_external'] = 'https://rwint-api-local.test/v1';
-$config['reliefweb_api.settings']['website'] = 'https://rwint-local.test';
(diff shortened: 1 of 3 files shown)


local/shared/settings/settings.local.php
```diff
@@ -177,3 +177,8 @@
   $settings['cache']['bins']['page'] = 'cache.backend.null';
   $settings['cache']['bins']['dynamic_page_cache'] = 'cache.backend.null';
 }
+
+$config['reliefweb_api.settings']['api_url'] = 'http://rwint-api-local-site/v1';
+$config['reliefweb_api.settings']['api_url_external'] = 'https://rwint-api-local.test/v1';
+$config['reliefweb_api.settings']['website'] = 'https://rwint-local.test';
+$config['reliefweb_api.settings']['elasticsearch'] = 'http://rwint-api-local-elasticsearch:9200';
```
orakili, line 184: This is already set here: https://github.com/UN-OCHA/rwint9-site/blob/baf3a5e90b5db9a953e510cb8214b70debca625a/local/shared/settings/settings.site.php.example#L27-L45.

The README mentions this file is to be renamed as part of the setup: https://github.com/UN-OCHA/rwint9-site/blob/baf3a5e90b5db9a953e510cb8214b70debca625a/local/README.md?plain=1#L8

lazysoundsystem: Right, I did have that set up right, the problem was I think in the elasticsearch indexing. I've removed the duplicated config.

orakili (approved):

PR #792, opened 2024-04-25
feat: Remove soft footer with mailchimp

attiks:
This PR removes the mailchimp signup form

Refs: #RW-945


lazysoundsystem (dismissed):
That should do it, but probably should remove
html/themes/custom/common_design_subtheme/templates/cd/cd-footer/cd-soft-footer.html.twig
html/themes/custom/common_design_subtheme/components/cd/cd-footer/cd-soft-footer.css
too, and the mention in 
html/themes/custom/common_design_subtheme/common_design_subtheme.libraries.yml

lazysoundsystem (approved):

PR #793, opened 2024-04-30
[RW-947] Blog subscriptions

orakili:
Refs: RW-947, RW-949, RW-950, RW-951, RW-952, RW-953

This PR adds a susbcription for the RW blog, adjusts the template to look similar to the mailchimp emails and adds a drush command to subscribe users for the transition.

There some new optional state variables to set for the deployment:

- `reliefweb_address`: physical address of RW, displayed, when set, at the bottom of the blog emails.
- `reliefweb_subscriptions_track_unsubscribe_links`: to enable/disable tracking parameters for the unsubscribe links, in case mail client do not like those for the "one-click" unsubscribe links
- `reliefweb_subscriptions_full_blog_post_copy`: enable the sending the full copy of the blog post instead of a summary

## Tests

Note: you need the standard local stack with a mailchimp instance

2. Log in and go to `/user/notifications`
3. Subscribe to the blog
4. Create a new blog post, publish it
5. Run `drush reliefweb_subscriptions:send` to send the queued notification of the published blog
6. Check local mailchimp for the sent email, check the email and various links it contains
7. Click the unsubscribe link at the bottom of the email, verify you have indeed been unsubscribe
8. Edit your account, take note of your email address
9. Run `echo "email.from@above" | drush -v -d reliefweb_subscriptions:subscribe-users blog`
10. Check that you're now subscribed again to the blog post
11. Go to `/blog`, check there is a "Subscribe" link that redirects you to your subscriptions page when logged in
12. Log out, repeat (11) and check that it redirects to the help page about subscriptions.


orakili:
@lazysoundsystem Looks like an index alias issue. The API uses the index aliases ex: `reliefweb_blog`.

`curl http://rwint-api-local-elasticsearch:9200/_cat/aliases` should give you the list of indices and their aliases.

Try to run `drush rapi-i --alias-only all` to fix the indices.

If that doesn't work, try a re-indexing with `drush rapi-i --tag=local2 --replace --alias all`.

lazysoundsystem (approved):
Thanks, yes, it was the aliases - using the second command fixed it all. I've noted it and will make sure to include it in the readme as I think it's hit me before.

I then couldn't see anything in mailpit, until I noticed the email needed verifying
rw-account-not-verified
but when I noticed that, everything worked exactly as described.

PR #795, opened 2024-04-30
Develop -> Main - v2.7.0

orakili:
## Features

- RW-885: one-click unsubscribe links for notifications
- RW-938: log manual changes by editors to content tagged by AI + command to send weekly stats

## Fixes

- RW-940: prevent parsing reference links when formatting revision logs
- OPS-9780: fix nginx rules to avoid hitting Drupal for files and images

## Chores

- OPS-9523: option to send emails via Amazon SES API instead of smtp
- RW-939: core patch to remove php notices
- RW-945: remove soft footer
- OPS-10304: adjust to latest coding standards
- composer update (minor updates)


html/modules/custom/reliefweb_reporting/src/Commands/ReliefWebReportingCommands.php
```diff
@@ -243,7 +244,7 @@ public function sendWeeklyReportStats($recipients) {
     $headers = [
       'From' => $from,
       'Reply-To' => $from,
-      'Content-Type' => 'multipart/mixed; boundary=GvXjxJ+pjyke8COw',
```
cafuego, line 246: Bye C0w 👋

PR #796, opened 2024-05-01
[RW-885] No need for personalized subscriptions page destination

orakili:
Refs: RW-885

The `/user/notifications` already redirects to the `/user/{uid}/notifications` page when logged in so let's avoid exposing the user ID unnecessarily.


cafuego (approved):
Both URLs work when logged in, but the latter does not leak user uid info.

PR #800, opened 2024-05-07
[RW-796] Allow use of the rich editor for reports

orakili:
Refs: RW-796

This sets the rich editor text format (ckeditor5 with some markdown conversion) as the default text format for reports similar to the jobs and training.

The default text format for existing content is still the `markdown` one to avoid losing data (the text can contain some extra stuff like image tracking code etc. tables which are really uncommon) but new reports use the `rich editor` text format.

To be able to switch between text formats without losing most of the content, this PR introduces some hack to convert the text between HTML (ckeditor 5) and markdown (textarea). See the comment in the `ReliefWebMarkdownXssFilter.php` file for mode details.

Jobs/training still continue to only use the Rich editor and should be unaffected by those changes (except for the addition of the blockquote option).

## Tests

**New report**

1. Log in as an editor
2. Create a report, confirm the body uses the `Rich editor` text editor
3. Add some content with bold, links etc.
4. Switch the text format to `markdown` and confirm the markdown text seems correct
5. Switch back to `Rich editor` and confirm the HTML is the same as (3)
6. Fill the rest of the mandatory fields
7. Save
8. Check the revision history at the bottom and confirm that the `body` is in markdown format
9. Edit again the report
10. Check that the text format is `Rich editor` and the body content seems correct
11. Switch to `markdown`
12. Change something in the title for example to see a new entry in the revision history
13. Save
14. Confirm that the revision history shows the same body (markdown) and that the content on the page is correct

**Old report**

1. Edit an old report (published before this PR)
2. Confirm that the text format is `markdown`
3. Save and confirm there is no new entry in the revision history or at least than the body is the same
4. Edit the report
5. Switch to `Rich editor`
6. Check that the content seems correct
7. Save, check the revision history has a new entry and that the markdown of the body is mostly the same as before

**Jobs/training**

1. Create a new job
2. Check that the body and how to apply use the rich editor **and** that there no text format selector
3. Fill in the form
4. Save, check that everything looks correct
5. Edit an existing job, change something in the body, save and check that everything is correct
6. Repeat the above for training

**Blog**

Edit a blog and confirm that only markdown or HTML are available as text formats (the rich editor may be added later after this PR is reviewed and validated).

### DIFF
--- a/config/core.entity_form_display.node.report.default.yml
+++ b/config/core.entity_form_display.node.report.default.yml
@@ -34,23 +34,22 @@ dependencies:
     - datetime
     - inline_entity_form
     - path
-    - reliefweb_files
     - reliefweb_fields
-    - text
+    - reliefweb_files
 id: node.report.default
 targetEntityType: node
 bundle: report
 mode: default
 content:
   body:
-    type: text_textarea_with_summary
+    type: reliefweb_formatted_text
     weight: 2
     region: content
     settings:
+      max_heading_level: 2
+      strip_embedded_content: true
       rows: 20
-      summary_rows: 3
       placeholder: ''
-      show_summary: false
     third_party_settings:
       allowed_formats:
         hide_help: '0'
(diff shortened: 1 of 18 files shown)


lazysoundsystem (approved):
Tested and everything works as described.

PR #810, opened 2024-05-16
[RW-975] Editor switch fix

orakili:
Refs: RW-975

Ensure the markdown version of a text field is saved when changing the editor from Rich editor to markdown

## Tests

1. Create a report, fill in the form (with some bold, links, list etc. in the body) and save
2. Edit the report, change the text format of the body to "Markdown" (**do not change the content of the body field**)
3. Save
4. Edit the report, the content of the body field is HTML instead markdown

1. Clear the cache
2. Repeat the steps above but this time (4) should show markdown instead of HTML in the body field


lazysoundsystem (approved):
Nice. Works as described.

PR #812, opened 2024-05-16
[RW-969] Post API - api key per user

attiks:
Still WIP

1. [x] Add a "trusted user" reference field to the provider entity type (filled by the RW team after verification/agreement etc. with partner).
2. [x] Add an API key field to users
3. [x] Add a user route for a user to get a new automatically generated key (view once)
4. [x ] Add a user route for a user to see the providers they are trusted for (so they can get the provider UUID notably)
5. [x] Update the provider entity to validate the API key against the API keys of the trusted users


orakili (commented):
Looks good, thanks.

PR #813, opened 2024-05-17
feat: Add seckit

attiks:
Refs: #OPS-10398

### DIFF
--- a/config/seckit.settings.yml
+++ b/config/seckit.settings.yml
@@ -1,0 +1,56 @@
+_core:
+  default_config_hash: x6bhN6WZwfVUI_LLMvRJIUW_2c26VTaBozbfXmJWmro
+seckit_xss:
+  csp:
+    checkbox: false
+    vendor-prefix:
+      x: false
+      webkit: false
+    report-only: false
+    default-src: ''
+    script-src: ''
+    object-src: ''
+    style-src: ''
+    img-src: ''
+    media-src: ''
+    frame-src: ''
+    frame-ancestors: ''
+    child-src: ''
+    font-src: ''
+    connect-src: ''
+    report-uri: /report-csp-violation
+    upgrade-req: false
+    policy-uri: ''
+  x_xss:
+    seckit_x_xss_option_disable: Disabled
+    seckit_x_xss_option_0: '0'
+    seckit_x_xss_option_1: 1;
+    seckit_x_xss_option_1_block: '1; mode=block'
+    select: 0
+seckit_csrf:
+  origin: false
+  origin_whitelist: ''
+seckit_clickjacking:
+  js_css_noscript: false
+  noscript_message: 'Sorry, you need to enable JavaScript to visit this website.'
+  x_frame: '1'
+  x_frame_allow_from: ''
+seckit_ssl:
+  hsts: false
+  hsts_subdomains: false
+  hsts_max_age: 1000
+  hsts_preload: false
+seckit_ct:
+  expect_ct: false
+  max_age: 86400
+  report_uri: ''
+  enforce: false
+seckit_fp:
+  feature_policy: false
+  feature_policy_policy: ''
+seckit_various:
+  from_origin: false
+  from_origin_destination: same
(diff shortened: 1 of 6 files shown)


lazysoundsystem (approved):
I tested this locally and found images from a number of OCHA domains blocked due to the img-src directive - as the pen-testing only objects to wildcards for styles and scripts and the list of domains for images was getting too long, I used a wildcard there.

I managed to deploy to sandbox by un-checking the 'Backup' option, https://reliefweb-sandbox.ahconu.org/. I have tested many different pages there without finding any CSP issues. More testing is required.

config/seckit.settings.yml
```diff
@@ -0,0 +1,56 @@
+_core:
+  default_config_hash: x6bhN6WZwfVUI_LLMvRJIUW_2c26VTaBozbfXmJWmro
+seckit_xss:
+  csp:
+    checkbox: true
+    vendor-prefix:
+      x: true
```
orakili, line 7: Is that necessary? This results in the addition of deprecated headers like `X-Content-Security-Policy-Report-Only` along the standard ones like `Content-Security-Policy-Report-Only`.

Ref: https://www.tenable.com/plugins/was/112552

orakili (approved):
I'm approving because I didn't find any CSP errors when reviewing on the sandbox site but, as mentioned in the comments below, there are extra stuff that I think could be removed.

config/seckit.settings.yml
```diff
@@ -0,0 +1,56 @@
...
+      webkit: false
+    report-only: false
+    default-src: "'self'"
+    script-src: "'self' 'unsafe-inline' 'unsafe-eval' fonts.googleapis.com www.gstatic.com https://*.google.com https://*.googletagmanager.com cdn.jsdelivr.net *.google-analytics.com https://cdnjs.cloudflare.com"
+    object-src: "'none'"
+    style-src: "'self' 'unsafe-inline' fonts.googleapis.com https://cdnjs.cloudflare.com"
+    img-src: "'self' data: https://*"
+    media-src: "'none'"
+    frame-src: "'self' https://*.mapbox.com"
```
orakili, line 16:
```suggestion
    frame-src: "'self' https://*.mapbox.com https://*.youtube.com https://*.youtu.be"
```
We need to add `https://*.youtube.com` and possibly `https://*.youtu.be`. For example for the video at the bottom of https://reliefweb.int/topic/iasc-task-force-5-localization-online-repository.

orakili (changes requested):

config/seckit.settings.yml
```diff
@@ -0,0 +1,56 @@
...
+  default_config_hash: x6bhN6WZwfVUI_LLMvRJIUW_2c26VTaBozbfXmJWmro
+seckit_xss:
+  csp:
+    checkbox: true
+    vendor-prefix:
+      x: true
+      webkit: false
+    report-only: false
+    default-src: "'self'"
+    script-src: "'self' 'unsafe-inline' 'unsafe-eval' fonts.googleapis.com www.gstatic.com https://*.google.com https://*.googletagmanager.com cdn.jsdelivr.net *.google-analytics.com https://cdnjs.cloudflare.com"
```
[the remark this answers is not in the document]
cafuego: It's *possible* they came in via New Relic, which we have since abandoned.

config/seckit.settings.yml
lazysoundsystem: If I'm reading this right, `X-Content-Security-Policy` is still used by IE, though `X-Webkit-CSP` isn't being used any longer.
https://caniuse.com/contentsecuritypolicy

This is complicated as https://caniuse.com/?search=contentsecuritypolicy-2 doesn't need any vendor prefixes, but the Seckit implementation doesn't use any of the new properties that CSP 2 includes (hashes, nonces, script-src-elem, etc)

orakili (approved):
Looks ok. I still think the vendor-prefix headers are not necessary and may be flagged when scanned but let's go with it.

PR #814, opened 2024-05-17
[RW-978] Fix: revert revision user for job tagger

orakili:
Refs: RW-978

**Note**: this a fix for the `v2.9.0` so the PR is against the `main` branch.

This revert the revision user change from https://github.com/UN-OCHA/rwint9-site/pull/802/commits/de10e81df6edf847d75fe775997ac13da41279f3. (The change for the timestamp is correct.)

Using the system user conflicts with https://github.com/UN-OCHA/rwint9-site/blob/56e52073fa5c7ce9211632c5fe0a39582d25959b/html/modules/custom/reliefweb_entities/src/OpportunityDocumentTrait.php#L21.


attiks (approved):

emmajane (approved):
🕹️ joysticking for monday in case that's useful.

PR #822, opened 2024-05-28
feat: Access check on reporting endpoint

attiks:


orakili (commented):
This seems to duplicate the access check done in the controller: https://github.com/UN-OCHA/rwint9-site/blob/fd7df8e1b29058519a34998cbecd6a65bf6596da/html/modules/custom/reliefweb_reporting/src/Controller/ReliefWebReportingController.php#L46

With the access check in the controller, the response if access is denied can still be a JSON response. With this new access class, an access denied returns the site's 403 page. Is it fine?

If yes, should the access check in the controller be removed as part of this PR?

cafuego:
Uh, what he said :-)

attiks:
Correct, access check can (and will) be removed from controller.

Regarding the format, it seems you need `jsonapi` to have support for `_format: 'json'`, there's a fix but it still depends on `jsonapi`, https://www.drupal.org/project/drupal/issues/3304981

I'll have a look

orakili (approved):
Working as expected. Thanks!

PR #826, opened 2024-06-04
chore: allow more google tracking

lazysoundsystem:
Refs: OPS-10398
OPS-10457

Note that this removes 'unsafe-eval', which I don't think we need. We should wait to merge this until we're sure we need the ad services, doubleclick, etc domains.


lazysoundsystem:
All of the google domains would have to be added for img-src too if we didn't have
the permissive `https://*` - wondering if that would be acceptable for
`connect-src` too - otherwise this is a very ugly header to send with
every request.

Also, presumably google updates their domain list every so often, so this list will go out of date.

orakili (approved):
Ouch, that's kind of insane.

lazysoundsystem:
Yes! what do you think of using `https://*` for connect-src too? If our xss protections are working correctly, it should be okay, shouldn't it?

orakili:
`https://*` is fine with me. It's slightly better than without the CSP.

orakili (approved):
Tested on `demo` and this seems to work.

lazysoundsystem:
Looks good to me too on demo. Merging.

PR #829, opened 2024-06-12
chore: show combined feedback stats and normalize scores

orakili:
Refs: RW-954

A proposal to show the combined stats and to normalize the scores. See the reasoning in the comment line 306-316.


attiks (approved):
👍

PR #831, opened 2024-06-13
Main -> Develop - v2.12.0

orakili:
## Features

- RW-954: job tagging test form + improvements to job tagging

## Chores

- OPS-10398: update CSP
- Composer update


config/ocha_ai.settings.yml
```diff
@@ -38,6 +38,17 @@ plugins:
       max_tokens: 8192
       api_secret: REPLACE_WITH_SECRET
       role_arn: null
+    aws_bedrock_titan_embed_text_v2:
```
cafuego, line 41: ah

jobtagger/tagjob.php
cafuego, line 1: I don't like this not being a drush command.

cafuego (approved):
Approved, but this will not work on prod until I add AWS settings for the model. Those are also not present on non-prod, so does that imply this isn't tested?

config/ocha_ai.settings.yml
cafuego: Ok, this uses the exiting bedrock settings so should be all good, eh?

jobtagger/tagjob.php
orakili: This is not going to be run normally. Pinging @attiks to see if that should be removed or refactored in a drush command later.

config/ocha_ai.settings.yml
orakili: Yes, it inherits the AWS settings from the parent plugin so there is nothing new to add.

jobtagger/tagjob.php
attiks: Was for testing only, I'll remove it

attiks: PR at https://github.com/UN-OCHA/rwint9-site/pull/835

PR #844, opened 2024-06-24
chore: include config changes with automatic updates

lazysoundsystem:
Refs: OPS-10254

### DIFF
--- a/.github/workflows/composer-update-config.yml
+++ b/.github/workflows/composer-update-config.yml
@@ -1,0 +1,20 @@
+name: Run Composer Update with Config
+
+on:
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Update The Thing with config
+      id: update-action-with-config
+      uses: UN-OCHA/actions/composer-update@OPS-10254-config-too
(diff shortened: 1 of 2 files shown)


lazysoundsystem:
This adds a new action for testing. Before replacing the weekly action, we'd need to merge https://github.com/UN-OCHA/actions/pull/37 - which seems to be fine for most repos but iis not yet working for Sesame.

orakili (approved):
Looks ok. @lazysoundsystem Is it the "final" version for this action or still a work in progress? (Can this be merged?)

lazysoundsystem:
Yes, it can be merged for testing. 

At the moment the action can only be triggered manually - when https://github.com/UN-OCHA/actions/pull/37 is merged, we can change the branch to main (like https://github.com/UN-OCHA/rwint9-site/blob/9156c7d3dba23fe175d390417b870feccb2573b7/.github/workflows/composer-update.yml#L14) and when we've tested that it works for RW we can add a scheduled trigger (like https://github.com/UN-OCHA/rwint9-site/blob/9156c7d3dba23fe175d390417b870feccb2573b7/.github/workflows/composer-update.yml#L4-L5)

PR #846, opened 2024-06-27
[RW-996] Chat public launch preparations

orakili:
Refs: RW-996

- [x] Allow anonymous users to access the AI chat
- [ ] Add honeypot to chat form
- [ ] Update `ocha_ai` to version with https://github.com/UN-OCHA/ocha_ai/pull/48


lazysoundsystem (approved):
Works as described. I wonder if we should extend this to check for language too? 
image

PR #848, opened 2024-07-02
[RW-1001] Handle dates equal or before unix epoch

orakili:
Refs: RW-1001

- Update datepicker version with fix
- Update different places handling publication dates

### Tests

**Note:** This requires a valid 

1. Checkout the branch, run composer install, clear the cache
2. Create a report, set the original publication date to 1 Jan 1970, check that this can be done via the datepicker
3. Check "bury"
4. Save
5. Go to `/updates`, then click "last" in the pager at the bottom and check that the document appears and 1 Jan 1970 is displayed for `posted` and `original publication date` <-- You need a valid installation of the API to test that (so the document is indexed)
6. Repeat 2 to 5 with a date before 1 Jan 1970, notably that the datepicker works

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -11098,16 +11098,16 @@
         },
         {
             "name": "reliefweb/simple-datepicker",
-            "version": "v1.3.3",
+            "version": "v1.3.4",
             "source": {
                 "type": "git",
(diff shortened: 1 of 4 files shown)


lazysoundsystem (approved):
This works well too.

PR #852, opened 2024-07-18
[RW-1005] AI chat logging, permission and settings

orakili:
Refs: RW-1005

- Install and set up jsonlog with patch to send data to ELK
- Update ocha_ai
- Update ocha_ai and ocha_ai_chat default settings
- Remove permission for authenticated users to use AI chat (will be enabled manually at first)


cafuego (approved):
Yup, that'll do. I am adding the required settings to ansible now.

PR #858, opened 2024-07-25
Ops 10638 weekly updates

lazysoundsystem:
Refs: OPS-10638

This moves to Drupal 10.3.

https://www.drupal.org/project/drupal/issues/3047110 has been merged, so the patch needed removing.

https://www.drupal.org/project/drupal/issues/3008292 is problematic - there's a new version of the patch on the issue, which didn't apply - the previous comment concludes the patch is no longer necessary. I haven't tested this yet.

https://www.drupal.org/project/imageapi_optimize_webp/issues/3453941 is needed to allow the site install. This patch has been added to all the other repos that are on 10.3.

Plus stage_file_proxy:^2 is no longer supported.


lazysoundsystem:
This has got messy - need to look more closely at the changelog. Very blocked.

lazysoundsystem:
@orakili - I was hoping to summarize this more cleanly before you get to updating core to 10.3, but some of this might be useful.

With these changes the automatic updates with config job runs (i.e. the site install part works which was blocked by the html/modules/custom/reliefweb_files/src/Controller/ImageStyleDownloadController.php change), but the tests are still failing and I haven't yet worked out why.

PR #863, opened 2024-07-29
feat: Index entities

attiks:
Refs: #RW-1035


cafuego (dismissed):
I think we need some extra settings in Ansible, eh? I am approving the PR, but:

* The lambda(s) should go into their own repository, which will also let us auto-deploy them.
* The README and code still think there are multiple datasource, knowlege bases and indices. Do we need another push? :-)

html/modules/custom/reliefweb_semantic/src/Commands/ReliefWebSemanticCommands.php
```diff
@@ -0,0 +1,630 @@
...
+        'created' => 'created',
+        'changed' => 'changed',
+        'title' => 'title',
+        'status' => 'status',
+        'body' => 'body',
+        'field_career_categories' => 'career_categories',
+        'field_city' => 'city',
+        'field_job_closing_date' => 'job_closing_date',
+        'field_country' => 'country',
+        'field_city' => 'city',
```
orakili, line 132: Duplicate of line 129

html/modules/custom/reliefweb_semantic/src/Commands/ReliefWebSemanticCommands.php
```diff
@@ -0,0 +1,630 @@
...
+        'uuid' => 'uuid',
+        'created' => 'created',
+        'changed' => 'changed',
+        'title' => 'title',
+        'status' => 'status',
+        'body' => 'body',
+        'field_country' => 'country',
+        'field_city' => 'city',
+        'field_source' => 'source',
+        'field_theme' => 'theme',
```
orakili, line 155: Probably need `field_training_type`, `field_training_language`, `field_fee_information` etc. (see /admin/structure/types/manage/training/fields)

cafuego (dismissed):
🕹️ for the cleanup. I have no opinions on the fields.

PR #864, opened 2024-07-30
feat: Embed using structured data

attiks:
Refs: #RW-1037

### DIFF
--- a/html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
+++ b/html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
@@ -414,7 +414,7 @@ function reliefweb_job_tagger_index_embedding(ContentEntityInterface $node) {
     'contents' => [
       'body' => [
         'type' => 'markdown',
-        'content' => $node->get('body')->value,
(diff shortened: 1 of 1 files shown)


html/modules/custom/reliefweb_job_tagger/reliefweb_job_tagger.module
```diff
@@ -400,8 +401,24 @@ function reliefweb_job_tagger_queue_job(Node $job) : void {
...
+  }
+
+  $data = [
+    'id' => $node->id(),
+    'title' => $node->get('title')->value,
+    'url' => $node->toUrl('canonical')->setAbsolute()->toString(),
+    'contents' => [
+      'body' => [
+        'type' => 'markdown',
+        'content' => $node->get('body')->value,
```
orakili, line 417: The title is not prepended here, which is different than what the RW source plugin does: https://github.com/UN-OCHA/ocha_ai/blob/a08ff287b3d326b8baa7914850f8efad295ee293/src/Plugin/ocha_ai/Source/ReliefWeb.php#L1088.

orakili (approved):

PR #865, opened 2024-08-05
[RW-1034] Drupal 10.3 upgrade

orakili:
Refs: RW-1034

- Bumps to PHP 8.3 with a few compatibility fixes
- Adds some coding standards fixes for the custom code
- Upgrades Drupal to 10.3.1 + dependency updates
- Removes unnecessary taxonomy patch
- Updates image upload validator core patch
- Adds Drupal 10.3 compatibility patches for image_optimize_webp
- Updates and syncs the config
- Updates the local test script to avoid email sending issue on site install


lazysoundsystem (approved):
Looks good - just one thing from other updates is that `stage_file_proxy` v2 is no longer supported (though the version here was only released a few weeks ago) and needs to be bumped to v3.

I spent a while trying to figure out the patch for https://www.drupal.org/project/drupal/issues/3008292 - the re-rolled patch in comment 104 didn't work for me, and 103 concludes it's no longer necessary - might be good to upload the patch there for the benefit of others.

orakili:
@lazysoundsystem Thanks, I'll update stage file proxy in a different PR.

As for the upload validator patch, I added it to the issue. It is still very much needed.

You can test it this way:

1. Build the site image **without** the patch
2. Create or find an image smaller than 64x64px
3. Create a topic, click add media -->notice that the minimum dimensions are not mentioned in the description
    Screenshot 2024-08-06 at 9 32 58
4. Upload your small image and notice it is accepted
5. Now, build the image with the patch
6. Create a topic, click add media --> the description mentions the minimum dimensions
    Screenshot 2024-08-06 at 9 36 38
7. Upload your small image and notice it is rejected

lazysoundsystem:
Thanks!

PR #866, opened 2024-08-06
feat: Show PHP (and Drupal) errors on fail.

cafuego:
Also show the docker version, for good measure.

Refs: OPS-10706

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -124,7 +124,7 @@ jobs:
         uses: cafuego/command-output@main
         with:
           run: |
-            docker compose -f tests/docker-compose.yml exec -T drupal drush -y si --existing-config minimal install_configure_form.enable_update_status_emails=NULL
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -153,6 +153,22 @@ jobs:
           threshold_metric: "lines"
           comment_footer: false
 
+      - name: DB Logs
+        id: dblog
+        if: failure()
+        uses: cafuego/command-output@main
+        with:
+          run: |
+            docker compose -f tests/docker-compose.yml exec -T drupal drush watchdog:show
```
orakili, line 162: I think that will return an error if, for example, the site install step failed, because the dblog module would not have been installed yet (line 136).

cafuego: Sure, but then the site install step failed, so we don't care about this one?

orakili: Will a failure of the DB log step prevent the display of the PHP fatal error in the next step?

cafuego: `fail-fast: true` is not set, so it should happily try to run the next step.

orakili (approved):

PR #869, opened 2024-08-15
feat: Social media links

attiks:
Refs: #RW-1049


html/modules/custom/reliefweb_entities/src/Entity/Source.php
```diff
@@ -207,27 +207,47 @@ protected function getOrganizationDetails() {
...
+          ];
+          break;
         }
       }
     }
 
     if (!empty($links)) {
+      uasort($links, function ($a, $b) {
+        return strcmp($a['title'], $b['title']);
+      });
```
orakili, line 249: I would suggest to use the collated sort from the LocalizationHelper: https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_utility/src/Helpers/LocalizationHelper.php#L82 to ensure it's in "natural order" and language aware.

```
LocalizationHelper::collatedSort($links, 'title');
```

lazysoundsystem (dismissed):

orakili (approved):

html/modules/custom/reliefweb_entities/src/Entity/Source.php
orakili: I made the change.

PR #885, opened 2024-09-12
chore: pass on credentials to action

lazysoundsystem:
Refs: OPS-10254

This is needed for the new version of the update action that includes config - https://github.com/UN-OCHA/rwint9-site/actions/runs/10825529694

### DIFF
--- a/.github/workflows/composer-update.yml
+++ b/.github/workflows/composer-update.yml
@@ -13,8 +13,10 @@ jobs:
       id: update-action
       uses: UN-OCHA/actions/composer-update@main
       with:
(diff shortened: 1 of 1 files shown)


lazysoundsystem:
To run the update, you'll need to remove the aggregation patch, e.g. https://github.com/UN-OCHA/response-site/pull/819

orakili (approved):

orakili:
@lazysoundsystem Can you precise what issue you have with the aggregation patch? It was pretty much needed to solve issues that prevented editors doing their job. The problem doesn't seem to have been solved in Drupal core. I can recreate it against Drupal 10.3.5 if the problem is that it doesn't apply correctly.

PR #900, opened 2024-09-24
[OPS-10834] EntraID login redirect

orakili:
Refs: OPS-10834

PR against https://github.com/UN-OCHA/rwint9-site/tree/cafuego/ops-10834-entraid-tweaks

- Return a 404 if the entraID config is not set properly (missing endpoints, for example on local)
- Update tests to check 404 and redirection
- Style user login links


cafuego (approved):
Bit of a rewrite more than a review eh ;-)

PR #911, opened 2024-10-02
[OPS-10917] Ensure we do not hit the rate limits for the tests or local devel

orakili:
Refs: OPS-10917


cafuego (approved):
Yup, that's how we do it :-) Though you're probably going to need a new laptop if you want to do 1k requests per second locally ;-)

PR #912, opened 2024-10-03
feat: UX of search convertor

attiks:
Refs: #RW-1086

image

### DIFF
--- a/html/modules/custom/reliefweb_rivers/js/search-converter.js
+++ b/html/modules/custom/reliefweb_rivers/js/search-converter.js
@@ -36,7 +36,7 @@
         // still showing feedback in case of failure to copy.
         el.addEventListener('mousedown', function (ev) {
           var tempInput = document.createElement('input');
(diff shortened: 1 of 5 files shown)


orakili (changes requested):
Looks good, thanks.

Just an issue: the copied text (ex: PHP) when pressing the copy button is HTML encoded:

```
[
  'query' =&gt; [
    'operator' =&gt; 'AND',
    'value' =&gt; 'source:OCHA',
  ],
  'offset' =&gt; 0,
  'limit' =&gt; 20,
  'preset' =&gt; 'latest',
  'profile' =&gt; 'list',
]
```

html/modules/custom/reliefweb_rivers/js/search-converter.js
```diff
@@ -0,0 +1,80 @@
+/* global once */
+(function () {
+  'use strict';
+
+  Drupal.behaviors.searchConvertor = {
```
orakili, line 5:
```suggestion
  Drupal.behaviors.searchConverter = {
```

orakili (approved):

PR #914, opened 2024-10-04
fix: Return the response object, so Drupal can do its thing.

cafuego:
Basically, this allows Drupal to initialize a session for the (anonymous) user so that the state token can be stored.

Refs: OPS-10834


cafuego:
This is hot-fixed on the dev site, so you can confirm this works.

orakili (approved):
Thanks!

PR #928, opened 2024-10-22
chore: update metatag module

lazysoundsystem:
Refs: OPS-10692


orakili (approved):
Checked several pages and there doesn't seem to be any changes to the meta tags.

PR #954, opened 2024-11-26
docs: enhance descriptions in api schema

lazysoundsystem:
Refs: RW-831

Includes the information from https://unitednations.sharepoint.com/:w:/r/sites/OCHAIMB/Digital%20Services%20Section/06_Projects/ReliefWeb/ReliefWeb%20API/ReliefWeb%20POST%20API%20-%20starting%202024/ReliefWeb%20Post%20API%20%E2%80%93%20Editorial%20guidelines.docx?d=w6774e15ece2349e0813973d88c6e01b4&csf=1&web=1&e=muNaiv - and tries to gather everything into one place.

I was looking for a reference for this, but couldn't find one:
```
                {
                    "description": "Must contain letters (any language).",
                    "pattern": "\\p{L}+"
                },
```
I understand that `L` also includes numbers, punctuation, html tags, etc, but I'm not sure what it excludes.

### DIFF
--- a/html/modules/custom/reliefweb_post_api/schemas/v2/report.json
+++ b/html/modules/custom/reliefweb_post_api/schemas/v2/report.json
@@ -45,7 +45,7 @@
             "maxItems": 30
         },
         "country": {
-             "description": "Document country(ies) as a list of IDs from https://api.reliefweb.int/v1/countries. The first country in the list is considered the primary country, meaning the most relevant to the content of the document. Otherwise, enter them alphabetically. Only tag the most relevant countries.",
+             "description": "Document country(ies) as a list of IDs from https://api.reliefweb.int/v1/countries. The first country in the list is considered the primary country, meaning the most relevant to the content of the document. Otherwise, enter them alphabetically. Only tag the most relevant countries as over-tagging often results in reduced discoverability.",
             "type": "array",
             "items": {
                 "type": "integer"
--- a/html/modules/custom/reliefweb_post_api/schemas/v2/report.json
+++ b/html/modules/custom/reliefweb_post_api/schemas/v2/report.json
@@ -242,7 +242,7 @@
             "required": ["url", "uuid", "checksum", "description", "copyright"]
         },
         "disaster": {
-            "description": "Document disaster(s) as a list of IDs from https://api.reliefweb.int/v1/disasters. Only select the most relevant terms for the document. Over-tagging often results in reduced discoverability.",
+            "description": "Document disaster(s) as a list of IDs from https://api.reliefweb.int/v1/disasters. Only select the most relevant disasters as over-tagging often results in reduced discoverability.",
             "type": "array",
             "items": {
                 "type": "integer"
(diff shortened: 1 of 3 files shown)


html/modules/custom/reliefweb_post_api/schemas/v2/training.json
```diff
@@ -54,7 +54,7 @@
             "maxItems": 2
         },
         "country": {
-            "description": "Training country(ies) as a list of IDs from https://api.reliefweb.int/v1/countries. Mandatory when the training format contains on-site otherwise not allowed.",
+            "description": "Training country(ies) as a list of IDs from https://api.reliefweb.int/v1/countries. Mandatory when the training format is 'on-site', otherwise not allowed.",
```
orakili, line 57: "is 'on-site'" is not correct because it is possible for a training to be both on-site and online thus the original "contains".

html/modules/custom/reliefweb_post_api/schemas/v2/report.json
```diff
@@ -242,23 +242,23 @@
             "required": ["url", "uuid", "checksum", "description", "copyright"]
         },
         "disaster": {
-            "description": "Document disaster(s) as a list of IDs from https://api.reliefweb.int/v1/disasters.",
+            "description": "Document disaster(s) as a list of IDs from https://api.reliefweb.int/v1/disasters. Only select the most relevant terms for the document. Over-tagging often results in reduced discoverability.",
```
orakili, line 245: "Only select the most relevant terms for the document." is only added here for the "disaster" field. Should this sentence be removed or added to the other taxonomy fields for consistency?

orakili (changes requested):

html/modules/custom/reliefweb_post_api/schemas/v2/training.json
```diff
@@ -231,7 +231,7 @@
     },
     "allOf": [
         {
-            "description": "The country field is mandatory if the training format contains on-site otherwise the country field is not allowed",
+            "description": "The country field is mandatory if the training format is 'on-site' otherwise the country field is not allowed",
```
orakili, line 234: Same as other comment about "on-site".

html/modules/custom/reliefweb_post_api/schemas/v2/report.json
lazysoundsystem: I think all the taxonomy terms that might be over-tagged are now covered

html/modules/custom/reliefweb_post_api/schemas/v2/training.json
lazysoundsystem: Good point - adjusted.

orakili:
@lazysoundsystem Could you also adjust the job and training schemas with sentences like "Only select the most relevant themes as over-tagging often results in reduced discoverability." for their taxonomy fields to be consistent with the report schema?

lazysoundsystem:
Thanks @orakili I've added to all fields that accept multiple options except countries and languages for jobs and training as it doesn't seem to apply to those.
By the way, for the training language, the API accepts up to 10 ids, but the guidelines say the Advertisement language can only be English, French or Spanish. 
Still looking for a reference for the `\L` character class for regular expressions - do you have something we can point to - or a more complete description?

orakili:
@lazysoundsystem See https://www.unicode.org/reports/tr44/#GC_Values_Table

orakili (approved):

lazysoundsystem:
Added a line in the apidoc linking to the unicode values table. Merging this now to include the updated schemas on the test site.

PR #965, opened 2024-12-10
fix: Tell composer to discard any changes we applied to vendor / contrib stuff via composer-patches.

cafuego:
Composer will re-apply the patches are update, so no need to make a mess by stashing or whatever.

This option is completely ignored when composer is in interactive mode.

Refs: RW-1128


cafuego:
This works now because https://github.com/UN-OCHA/actions/commit/04f5af701e0bf87c2b1f6971221e4952dd799ca7

orakili (approved):
Thanks!

PR #967, opened 2024-12-10
Update tag

attiks:


orakili:
Issue fixed by increasing the MySQL max connections (new MySQL image). See RW-1144. Closing this.

PR #968, opened 2024-12-11
Rw 1058 round 3

attiks:


html/modules/custom/reliefweb_entities/src/Services/ReportFormAlter.php
```diff
@@ -465,7 +465,6 @@ protected function alterFieldsForContributors(array &$form, FormStateInterface $
     // Hide fields.
     $form['field_bury']['#access'] = FALSE;
     $form['field_feature']['#access'] = FALSE;
-    $form['field_notify']['#access'] = FALSE;
```
orakili, line 468: Not seeing the field as a contributor. I think the template needs to be updated as well: https://github.com/UN-OCHA/rwint9-site/blob/e8b6ca0bc2b7d8fd4d5615d2385a5e2d6186ee21/html/themes/custom/common_design_subtheme/templates/form/node-edit-form--report.html.twig#L124-L131

html/modules/custom/reliefweb_fields/components/reliefweb-user-posting-rights/reliefweb-user-posting-rights.js
```diff
@@ -84,6 +84,43 @@
       return label;
     },
 
+    /**
+     * Create a filter for users.
+     */
+    createUserSelect: function () {
```
orakili, line 90: Not sure a user name `select` element is the best here. That doesn't allow to filter on the email address or user ID. It can also become difficult to use when there are many entries (50+) which was the complaint the editorial team has.

That said, they can use the built-in browser text search so maybe that's enough.

orakili (commented):
Missing access to the revision history (`view entity history` permission) for contributors.

html/modules/custom/reliefweb_fields/components/reliefweb-user-posting-rights/reliefweb-user-posting-rights.js
attiks: The idea was to have a textbox and not a select?

orakili: New filter is good, thanks.

orakili (approved):

orakili (approved):

PR #991, opened 2025-01-23
updates: weekly updates

lazysoundsystem:
Refs: OPS-11278

Note: I asked for an opinion on the updated amazon_ses patch on https://github.com/UN-OCHA/gms-site/pull/527


orakili:
@lazysoundsystem This patch is buggy, see https://github.com/UN-OCHA/gms-site/pull/527#issuecomment-2618124320. Closing this. I will create a PR with just the amazon_ses update and a new patch so we can run the composer update separately afterwards.

PR #998, opened 2025-01-28
chore: add csp exception for clarity analytics

lazysoundsystem:
Refs: RW-1164

The documentation suggests putting the exceptions for default-src, which
feels like overreach.


lazysoundsystem:
If this isn't included in the upcoming deployment - we can add the exceptions via the UI and merge this so they're not overwritten on the subsequent deployment.

orakili (approved):

PR #1030, opened 2025-03-04
[OPS-9523] Update and patch amazon_ses to refresh credentials

orakili:
Refs: OPS-9523

Amazon SES 3.1.1 with patch from https://www.drupal.org/project/amazon_ses/issues/3446425.


cafuego (approved):
Yessss.

PR #1054, opened 2025-04-15
feat: Import content from inoreader

attiks:
Refs: #RW-1192


attiks:
@orakili can you do a quick review, is anything else needed to trigger LLM or is all done by running `drush reliefweb-post-api:process`?

orakili:
@attiks The classification happens on cron runs but you can also run `drush queue:run ocha_classification_workflow:node_report` after `drush reliefweb-post-api:process`.

1. `drush reliefweb-post-api:process` --> creation of a report node --> queueing of the node for classification
2. `drush queue:run ocha_classification_workflow:node_report` or `drush cron` --> processing of the classification queue

The importer adds the items to the post api queue. I wonder if that's necessary as opposed to processing the items directly like the UNCHR and ECHO importers do. That may delay too much the creation of the reports. 

We could use the `$provider->skipQueue()` for the provider associated with the importer to determine whether to process directly or to queue. That would probably be the cleanest way.

Also we would need a hook implementation (ex: hook_entity_insert) to update the import record for items that were queued otherwise the import status will stay as "queued".

attiks:
See https://github.com/UN-OCHA/rwint9-site/pull/1070

PR #1060, opened 2025-04-17
fix: Ensure keyvalue.expire can detect credentials as expired during …

cafuego:
…a looooooong drush run.

Refs: OPS-9523


orakili (changes requested):
That will not change the REQUEST_TIME unfortunately. It's set once at the beginning.

Probably cleaner to load the profile and clear the KV store manually before recreating the client:

```
 /**
   * Get a new SesClient.
   *
   * This helps us if a mail run takes over an hour, in which case the
   * authentication token expires. This should generate us a new one
   * so we can continue.
   */
  protected function refreshClient() {
    // Delete the temporary credentials to ensure new ones are requested when
    // creating the new client.
    // @see Drupal\aws\Entity\Profile::getTemporaryCredentials()
    // @todo use proper dependency injection.
    $profile = \Drupal::service('aws')->getProfile('sesv2');
    if (isset($profile)) {
      \Drupal::service('keyvalue.expirable')->get('aws_profile')->delete($profile->id());
    }

    $this->client = $this->awsClientFactory->getClient('sesv2');
    $this->logger->info('Refreshing the SesV2 client after the session token expired');
  }
  ```

orakili (approved):

PR #1074, opened 2025-04-29
feat: Show the user a nicer error when uploading a giant file.

cafuego:
If the file is > the nginx max body size the user sees a bare nginx error page.  Override that, show an ajax in-line error if we can, and then redirect to a Drupal page that shows details if the user submits the form anyway.

Refs: OPS-11510


docker/etc/nginx/apps/drupal/drupal.conf
```diff
@@ -234,6 +234,26 @@ location @drupal-stage-file-proxy {
...
+
+   ## If callback, return in-line error box data.
+   if ($query_string ~ "ajax_form=1") {
+     add_header Content-Type application/json;
+     content_by_lua_file /etc/nginx/custom/lua/file_too_large_ajax.lua;
+   }
+
+   ## Override the request so we get a Drupal page.
+   fastcgi_param QUERY_STRING error=413&size=$content_length;
+   fastcgi_param REQUEST_URI /error/file-too-large;
```
cafuego, line 253: This page needs to exist on Drupal. It currently does on demo.

docker/etc/nginx/custom/lua/file_too_large_ajax.lua
```diff
@@ -0,0 +1,12 @@
+local cjson = require "cjson"
+
+local message = '<div class="" data-drupal-messages=""><div class="messages__wrapper"> <div class="messages messages--error cd-alert cd-alert--error" role="alert" data-drupal-message-type="error"> <div role="alert" aria-label="error"> <svg class="cd-icon cd-icon--error" aria-hidden="true" focusable="false" width="24" height="24"> <use xlink:href="#cd-icon--error"></use> </svg> <div class="cd-alert__container cd-max-width"> <div class="cd-alert__message [ cd-flow ]"> File Too Large. The file exceeds the maximum size of 25 MB. </div> <div class="cd-alert__title"> <h2 class="visually-hidden">Error message</h2> </div>  </div> </div> </div>'
```
orakili, line 3: Hmm, this hardcodes the max file size which depends on the version of the form (submitter = 25MB currently, probably will be reduced to fit the AWS Nova model limitations, 100MB for editors etc.).

Also the 413 can happened for different fields non only the file attachment one (ex: report image field).

Might be better to add a proper Drupal route with a controller and return the full page or the error code depending on whether its from ajax or not by checking Request::isXmlHttpRequest(). We could check if the current user is a submitter for example and retrieve the current file limit etc. and adjust based on the other info we can get from the request.

cafuego: Sure, but that's not MVP :-)

orakili: But nobody is using the document drop currently so no need to rush on that...

docker/Dockerfile
```diff
@@ -44,6 +44,8 @@ LABEL info.humanitarianresponse.build.date=$BUILD_DATE \
       info.humanitarianresponse.build.vcs-ref=$VCS_REF
 
 RUN mkdir -p /etc/nginx/custom && \
+    # Install the lua cjson library.
+    apk add -U --no-cache lua5.1-cjson && \
```
cafuego, line 48:
```suggestion
```

teodorescuserban (dismissed):

cafuego:
Ok. I think that *should* be the way translations get shipped.

attiks (dismissed):

html/modules/custom/reliefweb_utility/src/Controller/ErrorPageController.php
```diff
@@ -0,0 +1,148 @@
...
+
+      $response = [
+        [
+          'selector' => '#edit-' . strtr($field_name, ['_' => '-']),
+          'command'  => 'insert',
+          'method'   => 'append',
+          'data'     => '<div class="form-item--error-message">' . $message . '</div>',
+        ],
+      ];
+      return new JsonResponse($response, 200);
```
orakili, line 130: Unfortunately, this doesn't work for other fields like the report image field.

We cannot rely on `field_parents` to be set because it depends on the type of form element and type of field.

For example, for the report image field, there an `element_parents=field_image/widget/form/inline_entity_form/field_media_image/widget/0` but it's not quite the same and we cannot construct a reliable selector from it since it's the representation of the path in the `$form` array not directly used for the class, id or data-drupal-selector.

We also cannot retrieve the max file size for the field in a generic way. For this image field example, the max size is defined as the `field_media_image` level not the `field_image` level.

So I would propose to simply return a generic error message (same as the one line 136 to 141) at the top of the form just to have something and we'll add client side validation to handle the proper form error message in RW-1209:

```
use Drupal\Core\Ajax\AjaxResponse;
use Drupal\Core\Ajax\MessageCommand;

...

$response = new AjaxResponse();
$response->addCommand(new MessageCommand($message, NULL, ['type' => 'error']));
return $response;
```

html/modules/custom/reliefweb_utility/src/Controller/ErrorPageController.php
```diff
@@ -0,0 +1,148 @@
...
+    $message = $this->t("The file you are trying to upload is %size, which exceeds the maximum allowed size of %max_size. Please compress the file or choose a smaller one before trying again. If you need help reducing file size or have questions about upload limits, contact support: submit@reliefweb.int",
+      [
+        '%size' => $size,
+        '%max_size' => $max_size,
+      ]
+    );
+
+    return [
+      '#markup' => $message,
+    ];
```
orakili, line 145:
```suggestion
    return [
      '#type' => 'inline_template',
      '#template' => '<p class="rw-error-413">{{ message }}</p>',
      '#context' => ['message' => $message],
    ];
```

Just for minimum styling with some spacing between the title and message.

orakili (changes requested):

html/modules/custom/reliefweb_utility/src/Controller/ErrorPageController.php
cafuego: Hmm, I could have sworn I added a conditional for that. If it generally doesn't work though then yeah let's not bother.

cafuego: Also means the TypedDataManager is not needed anymore.

cafuego:
Now deployed as-is to the demo env. The reports page shows the generic error at the top of the page when a file is added and the generic error message page when submit is clicked.

It seems to not kick in when trying up upload a giant PNG file via the media library UI.

It does work as expected with the banner field on announcement pages.

teodorescuserban (approved):
nice :)

orakili (approved):

orakili:
> It seems to not kick in when trying up upload a giant PNG file via the media library UI.

Working for me for all upload fields, whether it's the media library UI (ex: blog post) or inline entity form (report image).

cafuego:
¯\_(ツ)_/¯

PR #1079, opened 2025-04-30
[RW-1216] Use timestamp of the most recent imported document as starting timestamp for the inoreader feeds

orakili:
Refs: RW-1216

Note: from my tests, `ot` accepts microsecond timestamps. That seems to be implied in https://www.inoreader.com/developers/stream-contents when they talk about the `timestampUsec` property.


attiks (approved):
Strange, but if it works, OK

I got 6 documents back from the API with the old code.

PR #1089, opened 2025-05-08
v2.30.0

orakili:
### Features

- OPS-11510: 413 file too large page
- RW-1214: auto-assign submitter role to UN people
- RW-1217: disable theme and disaster type tagging for inoreader imports

### Fixes

- OPS-9523: Delete cached SES credentials before recreating a client
- RW-1197: prevent PHP error when inoreader is not returning documents


cafuego (approved):
🕹️

PR #1096, opened 2025-05-12
Feature/rw 1207

attiks:


html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/WfpLogClusterImporter.php
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    // Retrieve the countries. Consider the first one as the primary country.
+    $countries = [];
+    foreach ($document['countries'] ?? [] as $location) {
+      if (isset($this->countryMapping[$location])) {
+        $country = $this->getCountryByIso($location);
+        $countries[] = $country;
+        $countries = array_unique($countries);
+      }
```
orakili, line 512:
```suggestion
      $country = $this->getCountryByIso($location);
      $countries[] = $country;
      $countries = array_unique($countries);
```

There is no `countryMapping` and the location is an ISO3 code.

config/reliefweb_import.plugin.importer.wfp_logcluster.yml
```diff
@@ -0,0 +1,18 @@
+enabled: 1
+max_import_attempts: '3'
+classification:
+  enabled: 1
```
orakili, line 4:
```suggestion
  enabled: false
```

Classification should be disable for the log cluster importer ?

orakili (changes requested):

config/reliefweb_import.plugin.importer.wfp_logcluster.yml
attiks: Yes

orakili (approved):

PR #1099, opened 2025-05-13
v2.31.0

orakili:
### Features

- RW-1207: Log Cluster importer
- RW-1218: Add option to skip last time stamp for inoreader importer
- RW-1227: Store file hash and add duplication detection based on it
- RW-1209: Client-side file size validation

### Fixes

- RW-1197: Make sure we have at least one file to import in inoreader
- RW-1224: Wrap long page titles
- RW-1197: Handle finding class among multiple classes in inoreader import tag


attiks:
Please add https://github.com/UN-OCHA/rwint9-site/pull/1104 as well

PR #1101, opened 2025-05-14
[RW-1207] Log cluster importer config change proposal

orakili:
Refs: RW-1207

@attiks This is relates to my questions in the [ticket](https://humanitarian.atlassian.net/browse/RW-1207?focusedCommentId=203530):

- Use `documents` endpoint instead of `activity-documents` to match the inoreader feeds and have documents sorted by date
- Reduce max-age to 3 days to avoid trying to import the same documents over and over
- Exclude "Meeting Minutes".


attiks:
Keeping meeting minutes

See https://github.com/UN-OCHA/rwint9-site/pull/1111

PR #1116, opened 2025-05-20
feat: Request PDF as blob

attiks:
Refs: #RW-1237


attiks:
Some notes:

1. PDF is return from the extract tool as a blob
2. Blob is saved locally
3. Code from ContentProcessor updated to allow `file://` protocol
4. Ideally we will save the PDF when calling `getRemoteFileInfo` so we don't have to download it twice.

html/modules/custom/reliefweb_import/reliefweb_import.install
```diff
@@ -143,3 +148,25 @@ function reliefweb_import_update_10002(array &$sandbox) {
...
+  $table_name = 'reliefweb_import_records';
+  $field_name = 'source';
+
+  if (!$schema->fieldExists($table_name, $field_name)) {
+    $table_schema = reliefweb_import_schema();
+    $schema->addField($table_name, $field_name, $table_schema[$table_name]['fields'][$field_name]);
+
+    // Update existing records.
+    \Drupal::database()->query("UPDATE {reliefweb_import_records} SET source = SUBSTR(SUBSTRING_INDEX(json_extract(extra, '$.inoreader.feed_name'), '[source:', 1), 2) where extra is not null");
+    \Drupal::database()->query("UPDATE {reliefweb_import_records} SET source = 'ECHO Flash Update' where importer = 'echo_flash_update'");
```
orakili, line 168:
```suggestion
    \Drupal::database()->query("UPDATE {reliefweb_import_records} SET source = 'ECHO Flash Update' where importer = 'echo_flash_update'");
    \Drupal::database()->query("UPDATE {reliefweb_import_records} SET source = 'ECHO Map' where importer = 'echo_map'");
```

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/EchoFlashUpdateImporter.php
```diff
@@ -419,16 +420,15 @@ protected function processDocuments(array $documents, string $provider_uuid, Con
           '@entity_id' => $entity->id(),
         ]));
       }
+      catch (DuplicateException $exception) {
+        $import_record['status'] = 'duplicate';
+        $import_record['message'] = $exception->getMessage();
+        $import_record['attempts'] = $max_import_attempts;
```
orakili, line 426:
```suggestion
        $import_record['attempts'] = $max_import_attempts;
        $this->getLogger()->error(strtr('Unable to process @source document @id: @exception', [
          '@source' => $this->sourceName,
          '@id' => $id,
          '@exception' => $exception->getMessage(),
        ]));
```

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/InoreaderImporter.php
```diff
@@ -387,6 +388,7 @@ protected function processDocuments(array $documents, string $provider_uuid, Con
         'status' => 'pending',
         'message' => '',
         'attempts' => 0,
+        'source' => trim(substr($document['origin']['title'] ?? '', 0, strpos($document['origin']['title'] ?? '', '[source:'))),
```
orakili, line 391:
```suggestion
        'source' => trim(substr($document['origin']['title'] ?? '', 0, strpos($document['origin']['title'] ?? '', '[source:') ?: NULL)),
```

Use the full title if `[source:` could not be found.

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/InoreaderImporter.php
```diff
@@ -511,16 +513,15 @@ protected function processDocuments(array $documents, string $provider_uuid, Con
           '@entity_id' => $entity->id(),
         ]));
       }
+      catch (DuplicateException $exception) {
+        $import_record['status'] = 'duplicate';
+        $import_record['message'] = $exception->getMessage();
+        $import_record['attempts'] = $max_import_attempts;
```
orakili, line 519:
```suggestion
        $import_record['attempts'] = $max_import_attempts;
        $this->getLogger()->error(strtr('Unable to process Inoreader @id: @exception', [
          '@id' => $id,
          '@exception' => $exception->getMessage(),
        ]));
```

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/UnhcrDataImporter.php
```diff
@@ -1005,16 +1006,15 @@ protected function processDocuments(array $documents, string $provider_uuid, Con
           '@entity_id' => $entity->id(),
         ]));
       }
+      catch (DuplicateException $exception) {
+        $import_record['status'] = 'duplicate';
+        $import_record['message'] = $exception->getMessage();
+        $import_record['attempts'] = $max_import_attempts;
```
orakili, line 1012:
```suggestion
        $import_record['attempts'] = $max_import_attempts;
        $this->getLogger()->error(strtr('Unable to process UNHCR document @id: @exception', [
          '@id' => $id,
          '@exception' => $exception->getMessage(),
        ]));
```

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/WfpLogClusterImporter.php
```diff
@@ -477,16 +478,15 @@ protected function processDocuments(array $documents, string $provider_uuid, Con
           '@entity_id' => $entity->id(),
         ]));
       }
+      catch (DuplicateException $exception) {
+        $import_record['status'] = 'duplicate';
+        $import_record['message'] = $exception->getMessage();
+        $import_record['attempts'] = $max_import_attempts;
```
orakili, line 484:
```suggestion
        $import_record['attempts'] = $max_import_attempts;
        $this->getLogger()->error(strtr('Unable to process WFP Logcluster document @id: @exception', [
          '@id' => $id,
          '@exception' => $exception->getMessage(),
        ]));
```

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/InoreaderImporter.php
```diff
@@ -549,7 +550,10 @@ protected function processDocumentData(string $uuid, array $document): array {
     $url = $document['canonical'][0]['href'];
 
     // Retrieve the title and clean it.
-    $title = $this->sanitizeText(html_entity_decode($document['title'] ?? ''));
+    $title = $this->sanitizeText(html_entity_decode($document['title'] ?? ''), TRUE);
```
orakili, line 553: Why the `TRUE`? This will preserve new lines but that's not something we allow for titles usually.

orakili (commented):
Rather than using `file://` and saving the file temporarily, I think it might be better to allow the post api content processor to receive raw bytes.

This way, we may be able to avoid double download (first in the imported to retrieve the checksum and then in the post api processor) for normal files with a URI as well as this case where the blob is retrieved with puppeteer without using an intermediary file.

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/InoreaderImporter.php
attiks: Fixed

PR #1133, opened 2025-06-03
feat: Add worldbank API

attiks:


config/reliefweb_import.plugin.importer.worldbank.yml
```diff
@@ -0,0 +1,19 @@
+enabled: true
+max_import_attempts: 3
+classification:
+  enabled: true
```
orakili, line 4: Do we need the AI classification? Isn't the data from the API good enough?

orakili (approved):

config/reliefweb_import.plugin.importer.worldbank.yml
attiks: Discussed with Mary, using AI for theme and content format

orakili (dismissed):

PR #1139, opened 2025-06-11
[RW-1202][RW-1203][RW-1252] UX improvements for submitters (mostly)

orakili:
Refs: RW-1202, RW-1203, RW-1252

- Add copy document link button to reports, jobs, training and blog posts below social media links
- Change the message after submitting node form to indicate if published or unpublished and add copy link button in message
- Show published/unpublished as moderation status on submitter content


attiks (approved):
Looks good

PR #1145, opened 2025-06-13
feat: Add social_auth with support for Google, LinkedIn and Microsoft.

cafuego:
Refs: OPS-11576 OPS-11600


composer.json
```diff
@@ -52,7 +52,10 @@
         "drupal/pathauto": "^1.11",
         "drupal/redirect": "^1.8",
         "drupal/seckit": "^2.0",
-        "drupal/social_auth_hid": "^3.2",
+        "drupal/social_auth_google": "^4",
+        "drupal/social_auth_hid": "^5",
+        "drupal/social_auth_linkedin": "^4",
+        "drupal/social_auth_microsoft": "^4",
```
orakili, line 58: Is that still used? The config was removed in https://github.com/UN-OCHA/rwint9-site/pull/1145/commits/73ebd0be4fae2903047124b0e597913dfacdf383

config/core.extension.yml
```diff
@@ -101,7 +101,10 @@ module:
   seckit: 0
   social_api: 0
   social_auth: 0
+  social_auth_google: 0
   social_auth_hid: 0
+  social_auth_linkedin: 0
+  social_auth_microsoft: 0
```
orakili, line 107: Is that still used? The config was removed in https://github.com/UN-OCHA/rwint9-site/commit/73ebd0be4fae2903047124b0e597913dfacdf383

cafuego: It is it, it was. Turns out the that drush will not validate the config if the file is present; it says it should not exist.

In its defence, if I enable the module, I cannot export social_auth_microsoft config. The config overrides via settings.php work fine though, so I'm going to guess that maybe the schema is missing or broken.

cafuego: Yes, but it still errored out when the name was correct 😢 - https://github.com/UN-OCHA/rwint9-site/actions/runs/15625868466/job/44019892443

orakili: With `config/social_auth_microsoft.settings.yml`, the tests seem to pass: https://github.com/UN-OCHA/rwint9-site/pull/1146

orakili (approved):

config/core.extension.yml
cafuego: :face palm:

PR #1153, opened 2025-06-18
test

orakili:


orakili:
> @orakili Yeah, fine now. The issue was a bad xdebug.

Thanks!

PR #1157, opened 2025-06-20
chore: Remove the text underline from the login button-ish links by popular request.

cafuego:
Colour changes, tiny icon images and other style changes are likely to follow in the not too distant future.


orakili (commented):
We lose on accessibility by removing the underline because then only the color indicates the state change (hovered etc.).

So I would propose to either underline when hovered/active etc. and remove from the initial state or keep the current version.

cafuego:
Changed the PR to keep the underline on hover/active. Mind you this is in response to a question on the button styling on the daily call, not an actual ticket.

PR #1159, opened 2025-06-23
feat: FAO data from API

attiks:
Refs: #RW-1245

### DIFF
--- a/html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
+++ b/html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
@@ -453,6 +453,20 @@ class FaoDataImporter extends ReliefWebImporterPluginBase {
         }
         break;
 
+      case 'Microsoft Excel':
+        if (isset($document['pdfLink'])) {
+          $info = $this->getRemoteFileInfo($document['pdfLink'], 'xlsx');
+          if (!empty($info)) {
+            $file_url = $document['pdfLink'];
+            $file_uuid = $this->generateUuid($file_url, $uuid);
+            $files[] = [
+              'url' => $file_url,
+              'uuid' => $file_uuid,
+            ] + $info;
+          }
+        }
+        break;
+
       default:
         return [];
 


html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
```diff
@@ -0,0 +1,466 @@
...
+        '@id' => $id,
+      ]));
+
+      // Generate a hash from the data we use to import the document. This is
+      // used to detect changes that can affect the document on ReliefWeb.
+      $filtered_document = $this->filterArrayByKeys($document, [
+        'id',
+        'title',
+        'created',
+        'modified',
```
orakili, line 282: I think modified should not be included in the data used to generate the hash because this may be updated for properties that should not trigger a re-import.

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
```diff
@@ -0,0 +1,466 @@
...
+        'documentLink',
+      ]);
+      $hash = HashHelper::generateHash($filtered_document);
+      $import_record['imported_data_hash'] = $hash;
+
+      // Legacy hash.
+      // @todo possibly remove in a few months (now is 2025-04-21).
+      // @see RW-1196
+      $legacy_hash = HashHelper::generateHash($document, ['updated', 'downloadCount']);
```
orakili, line 291: New importer, no legacy hash.

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
```diff
@@ -0,0 +1,466 @@
...
+
+      // Skip if there is already an entity with the same UUID and same content
+      // hash since it means the document has been not updated since the last
+      // time it was imported.
+      $records = $this->entityTypeManager
+        ->getStorage($entity_type_id)
+        ->getQuery()
+        ->accessCheck(FALSE)
+        ->condition('uuid', $uuid, '=')
+        ->condition('field_post_api_hash', [$legacy_hash, $hash], 'IN')
```
orakili, line 301: no legacy hash

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
```diff
@@ -0,0 +1,466 @@
...
+      $info = $this->getRemoteFileInfo($document['pdfLink']);
+      if (!empty($info)) {
+        $file_url = $document['pdfLink'];
+        $file_uuid = $this->generateUuid($file_url, $uuid);
+        $files[] = [
+          'url' => $file_url,
+          'uuid' => $file_uuid,
+        ] + $info;
+      }
+    }
```
orakili, line 423: Bail out if the file could not be retrieved?

html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporter/FaoDataImporter.php
```diff
@@ -0,0 +1,466 @@
...
+  protected function getDocuments(int $limit, string $order_property = 'created'): array {
+    // Get list of documents.
+    try {
+      $timeout = $this->getPluginSetting('timeout', 5, FALSE);
+      $api_url = $this->getPluginSetting('api_url');
+
+      // Query the FAO API.
+      $query = http_build_query([
+        'limit' => $limit,
+        'type' => 'PDF',
```
orakili, line 117: Looks like we could use something like `filter=(type=PDF AND (created BETWEEN 1745938800000 AND 1750517999999))` to filter for PDF and restrict the date range. So we could store the timestamp of the latest import and use that for the lower time limit to reduce import attempts of the same content. What do you think?

attiks: :confused: code updated, will generate a link like `https://data-in-emergencies.fao.org/api/search/v1/collections/document/items?limit=5&sortBy=%2Bproperties.created&filter=%28type%3DPDF+AND+%28created+BETWEEN+1650168396+AND+1750754796%29%29` but this doesn't return anything

I tested on their site as well, but filter doesn't apply

orakili: The timestamps need to be in milliseconds. `https://data-in-emergencies.fao.org/api/search/v1/collections/document/items?limit=5&sortBy=%2Bproperties.created&filter=(type%3DPDF+AND+(created+BETWEEN+1650168396000+AND+1750754796000))` seems to work.

attiks: Seems I needed to go back in time a bit more

:+1:

orakili:
Closing as per comment on ticket.

PR #1164, opened 2025-06-25
feat: Pretty login buttons a-la sample screenshot.

cafuego:
It's at least a place to start. Not sure if the `UN Agency` one should be styled the same, but I haven't dropped in an override and/or logo for that as yet.

Ninja-added on https://demo.reliefweb-int.ahconu.org/user/login


html/themes/custom/common_design_subtheme/components/rw-document/rw-document.css
```diff
@@ -129,22 +129,49 @@
...
-  border-radius: 3px;
+  border-radius: 2rem;
   background: var(--cd-reliefweb-brand-blue--dark);
   font-weight: bold;
   text-decoration: none;
 }
+
+.rw-document.rw-article--book .rw-article__content a[href^="/user/login/"]::before {
+  position: relative;
+  top: 1px;
```
cafuego, line 144:
```suggestion
  top: 2px;
```

Hmmm. On closer inspection the ::before top should probably be 2px and not 1px so as to not offend my eyes :-)

orakili (approved):
The logos could use a pass at https://jakearchibald.github.io/svgomg/ but otherwise good.

orakili (dismissed):

orakili (approved):

PR #1167, opened 2025-06-25
feat: Add editorial flow

attiks:
Refs: #1273


html/modules/custom/reliefweb_import/src/Service/ReliefWebImporterModeration.php
```diff
@@ -280,15 +282,53 @@ public function getRows(array $results) {
...
+        }
+
+        $cells['node_created']['editor_label'] = [
+          '#type' => 'markup',
+          '#markup' => $this->t('Workflow'),
+        ];
+        $cells['node_created']['editor'] = [
+          '#type' => 'dropbutton',
+          '#dropbutton_type' => 'rw-moderation',
+          '#links' => $editor_links,
```
orakili, line 322: Should that be changed to editorial_flow here as well?

orakili (approved):

PR #1188, opened 2025-07-11
[RW-1294] Change logic for origin filter to properly respect the posted roles hierarchy

orakili:
Refs: RW-1294


orakili:
Merged and deployed in v2.38.1

PR #1189, opened 2025-07-13
v2.38.1

orakili:
### Fixes

- RW-1294: change logic for origin filter to properly respect the poster roles hierarchy


cafuego (approved):
Severe #yolo review though

PR #1192, opened 2025-07-15
v3.0.0

orakili:
### Features

- OPS-11475, RW-1286: Drupal 11 upgrade


cafuego (approved):
Passes tests, loads OK and seems to not cause more errors in ELK than the D10 version, so I think that's a 👍

PR #1198, opened 2025-07-18
chore: composer update

orakili:
https://github.com/UN-OCHA/rwint9-site/actions/runs/16334324496 failed but manual composer update worked.


cafuego (approved):
Hmm. Did you run without `--no-scripts` locally maybe?

orakili:
> Hmm. Did you run without `--no-scripts` locally maybe?

That's correct. I get the same error with the `--no-scripts` since it doesn't execute the lenient script that is necessary for modules which are still stuck with Drupal 10 requirement.

orakili:
Closing in favor of https://github.com/UN-OCHA/rwint9-site/pull/1200#pullrequestreview-3040433128

PR #1222, opened 2025-08-12
feat: Import organizations

attiks:
Refs: #RW-1332

### DIFF
--- a/config/core.extension.yml
+++ b/config/core.extension.yml
@@ -39,6 +39,7 @@ module:
   inline_entity_form: 0
   inline_form_errors: 0
   jsonlog: 0
+  jsonrpc: 0
   key: 0
   language: 0
   link: 0
@@ -91,6 +92,7 @@ module:
   reliefweb_revisions: 0
   reliefweb_rivers: 0
   reliefweb_subscriptions: 0
+  reliefweb_sync_orgs: 0
   reliefweb_topics: 0
   reliefweb_user_history: 0
   reliefweb_user_posts: 0
--- a/config/user.role.webmaster.yml
+++ b/config/user.role.webmaster.yml
@@ -33,6 +33,7 @@ dependencies:
     - reliefweb_guidelines
     - reliefweb_moderation
     - reliefweb_post_api
+    - reliefweb_sync_orgs
     - reliefweb_users
     - system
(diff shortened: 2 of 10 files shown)


attiks:
Modification are needed since `$updated_fields` is set on [line 245](https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_ai/reliefweb_ai.module#L245) and is set again on [line 254](https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_ai/reliefweb_ai.module#L254) so the returned value doesn't contain all changed fields.

I'll add a new hook

attiks:
> > Modification are needed since `$updated_fields` is set on [line 245](https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_ai/reliefweb_ai.module#L245) and is set again on [line 254](https://github.com/UN-OCHA/rwint9-site/blob/develop/html/modules/custom/reliefweb_ai/reliefweb_ai.module#L254) so the returned value doesn't contain all changed fields.
> > I'll add a new hook
> 
> Ah, I had another look and actually the logic is wrong in the `reliefweb_ai_ocha_content_classification_post_classify_entity` function. It should be:
> 
> ```
>   // Attempt to retrieve the language.
>   $language_updated_fields = reliefweb_ai_post_classification_language_update(
>     $entity,
>     $workflow,
>     $classifier,
>     $updated_fields,
>     $data,
>   ) ?? [];
> 
>   // Attempt to fix the body.
>   $body_updated_fields = reliefweb_ai_post_classification_body_update(
>     $entity,
>     $workflow,
>     $classifier,
>     $updated_fields,
>     $data,
>   ) ?? [];
> 
>   return array_unique(array_merge($language_updated_fields, $body_updated_fields)) ?: NULL;
> ```
> 
> So that the returned value is either the list of extra changed fields (language, body) or NULL if no changed occurred.
> 
> This is a logic mistake from when I added the body fix (`reliefweb_ai_post_classification_body_update`).
> 
> The output of the post_classify hook implementations is expected to be an array with only the fields that were updated in that implementation or NULL if none were updated.

I'll update the code and revert my changes

PR #1226, opened 2025-08-15
fix: Check if there is an origin before trying to use it.

cafuego:
NOTICE: PHP message: Uncaught PHP Exception Error: "Call to a member function getPossibleOptions() on null" at /srv/www/html/modules/custom/reliefweb_moderation/src/Services/ReportModeration.php line 212

### DIFF
--- a/html/modules/custom/reliefweb_moderation/src/Services/ReportModeration.php
+++ b/html/modules/custom/reliefweb_moderation/src/Services/ReportModeration.php
@@ -209,20 +209,23 @@ class ReportModeration extends ModerationServiceBase {
       $cells['data'] = array_filter($data);
 
       // Retrieve the origin of the document.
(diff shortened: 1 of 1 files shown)


cafuego:
There are more, but they are exceptions thrown in core funxtion and I have no debug traces to see where they came from.

```
Uncaught PHP Exception TypeError: "Drupal\Component\Utility\Html::escape(): Argument #1 ($text) must be of type string, null given, called in /srv/www/html/core/lib/Drupal/Core/Utility/LinkGenerator.php on line 199" at /srv/www/html/core/lib/Drupal/Component/Utility/Html.php line 433
```

```
Uncaught PHP Exception Exception: "Only file JavaScript assets with preprocessing enabled can be optimized." at /srv/www/html/core/lib/Drupal/Core/Asset/JsOptimizer.php line 34"
```

orakili (approved):

PR #1372, opened 2026-03-19
[RW-1470] Do not discard valid most-read entries

orakili:
Refs: RW-1470


cafuego (approved):
It it's good enough to hack live on prod, it's good enough to merge.

PR #1388, opened 2026-04-09
[RW-1476] Remove ending slash to be compatible with API gateway change for log cluster API

orakili:
Refs: RW-1476


cafuego (approved):
(╯°□°）╯︵ ┻━┻

PR #1410, opened 2026-05-11
chore: update patch for pathauto 8.x-1.15

orakili:
Update pathauto and its patch to fix composer update github action.


cafuego (approved):
Yup.

Note there is also https://www.drupal.org/i/3579655 with https://git.drupalcode.org/project/pathauto/-/merge_requests/173/diffs

orakili:
Yes, I saw that but the patch in that MR is way bigger for no added benefit.

PR #1424, opened 2026-05-21
fix(security): Mitigate an nginx CVE in regex rewrites.

cafuego:
Use named captures everywhere for values that should be re-emitted.

Refs: OPS-12132


orakili (approved):
Thanks!

PR #1428, opened 2026-06-02
[RW-1475] Report series matching service

orakili:
Refs: RW-1475

This contains:

### Series detection (RW-1475)
- Report series matching service
- Flag to skip API indexing when an entity is saved
- Publication notification moved to their own service
- ID list parameter for the drush command to download missing files
- New drush command to generate missing file previews

### Duplicate detection (RW-1484)
On report creation, this looks for duplicate document (similar body text) from the same source(s) posted within a week and mark the document as duplicate or to review.

When available, this extends the search for duplicates using embeddings to find same documents posted by other sources or in the past.

This requires version 1.2.0 of the OCHA AI Helper for the embedding endpoints. Currently it uses MariaDB VECTOR functionalities to store the embeddings because it's fast, doesn't require another store and allows to optimize filtering without duplicating data. This could be replaced with a different backend by implementing a new backend service.


html/modules/custom/reliefweb_files/src/Commands/ReliefWebFilesCommands.php
```diff
@@ -500,13 +517,25 @@ public function downloadMissingFiles(
     $chunk_size = $options['chunk-size'] ?? 50;
     $timeout = $options['timeout'] ?? 30;
 
+    $node_ids = [];
+    if (!empty($options['ids'])) {
+      $node_ids = array_values(array_unique(array_filter(
```
cafuego: This is the most ridiculous set of nested array functions I have ever seen 😂

orakili: Yeah 😢 but they all have a job to do. You have the same one line 860. This was short and effective so for a drush command used for development 🤷

cafuego: So apparently there's a way to make this slightly nicer (for some specific values of "nicer") in PHP 8.5 - the pipe operator.

```
$node_ids = $options['ids']
  |> fn($v) => explode(',', $v)
  |> fn($v) => array_map('trim', $v)
  |> fn($v) => array_map('intval', $v)
  |> array_filter(...)
  |> array_unique(...)
  |> array_values(...);
```

cafuego (dismissed):
The parts I understand look good :-)

cafuego (dismissed):

cafuego (approved):
Re-approving.

I just had a quick look over the code again; there are a bunch of magic numbers in there for the score. If we need or want to tweak those, we'll have to do a deploy. Might be worth looking at changing them all to a bunch of config/settings if it turns out they *do* need tweaking in the future.

orakili:
All numbers used in the calculations are now set via the configuration.

cafuego (approved):

PR #1436, opened 2026-06-17
[RW-1421] Use enum for posting rights to be less error prone

orakili:
Refs: RW-1421

This is a refactor to use an enum instead of hardcoded strings and ints. Functionality should not be changed.


cafuego (dismissed):
I did a bit of twiddling locally and the logic appears sound. I did not spin up a whole instance and test on actual content.

orakili:
Tiny revert to fix a changed behavior during the refactor for edge cases.

cafuego (approved):
Oh I don't know.

PR #1446, opened 2026-07-02
fix(build): Allow the symfony runtime plugin to run.

cafuego:
Refs: OPS-12194


orakili (dismissed):

orakili:
@cafuego I marked this PR as `blocked` so that we do not merge it before the deployments to convert the guidelines otherwise there will too many moving pieces. I'll do the deployments early next week and then proceed with the upgrade to Drupal 11.4 from this PR and https://github.com/UN-OCHA/rwint9-site/pull/1447.

orakili (approved):

PR #1453, opened 2026-07-07
[RW-1515] Fix guideline sort form local task preventing term access

orakili:
Refs: RW-1515

This relaxes the type checking so that the guideline sort form local task doesn't cause an error preventing access to all terms...


cafuego (approved):
Tested on prod and working well.

PR #1458, opened 2026-07-10
[RW-1514] Strengthen guideline access

orakili:
Refs: RW-1514

Improve guideline access check with better permissions and service.


cafuego (changes requested):
Failed asserting that false is true :-)

orakili:
OK, now true is true and false is false.

cafuego (approved):

PR #1469, opened 2026-07-23
[RW-1522] Patch amazon_ses to handle quoted names with commas

orakili:
Refs: RW-1522


cafuego (approved):
Wheeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #1476, opened 2026-07-30
feat: Use AWS iam role auth, so we do not need to roll credentials.

cafuego:
Refs: OPS-12073

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,6 +8,7 @@ jobs:
     permissions:
       contents: write
       checks: write
+      id-token: write
       pull-requests: write
       actions: read
       statuses: write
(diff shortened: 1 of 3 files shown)


cafuego:
The actions already require the new auth vars, so this needs to be merged for any builds to work.

orakili (dismissed):

.github/workflows/run-tests.yml
```diff
@@ -83,6 +83,7 @@ jobs:
         id: aws-login
         uses: aws-actions/amazon-ecr-login@v2
         if: ${{ !env.ACT }}
+        with:
```
cafuego, line 86: This is the problem with making the same change to 15 repos. Eventually you mess it up and don't even notice 😭

orakili (approved):

PR #1480, opened 2026-08-04
[RW-1527] Add configurable file download throttling and user agent for importers

orakili:
Refs: RW-1527

Note: the user agents will have to be set separately.

### DIFF
--- a/html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporterPluginBase.php
+++ b/html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporterPluginBase.php
@@ -35,6 +35,7 @@ use Drupal\reliefweb_post_api\Plugin\ContentProcessorPluginManagerInterface;
 use Drupal\reliefweb_utility\Helpers\TextHelper;
 use Drupal\reliefweb_utility\Helpers\UrlHelper;
 use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\TransferException;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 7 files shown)


html/modules/custom/reliefweb_import/src/Plugin/ReliefWebImporterPluginBase.php
```diff
@@ -1316,6 +1283,161 @@ protected function getRemoteFileInfo(string $url, string $default_extension = 'p
...
+    // Bot/WAF filtering: try the next user agent.
+    if ($status === 403) {
+      $this->getLogger()->notice(strtr('Bot/WAF filtering on attempt for @url with user agent "@agent".', [
+        '@url' => $url,
+        '@agent' => $user_agent,
+      ]));
+      return [$response, TRUE];
+    }
+
+    // Other error: do not keep trying.
```
cafuego, line 1448: Let's log errors, though :-)

```suggestion
    // Other error: do not keep trying.
    $this->getLogger()->notice(strtr('HTTP @status error on attempt for @url with user agent "@agent".', [
      '@status' => $status,
      '@url' => $url,
      '@agent' => $user_agent,
    ]));
```

cafuego (dismissed):
Hmm... I hate bots that rotate user-agents, and I block them when I can :-)

In theory this change is OK. The extra logging suggestion I added isn't necessary I guess given the error thrown upstream; although in that case the URL isn't logged.

Not sure if you want to make a distinction between client side errors and server side errors, if only for ease of log filtering on ELK.

cafuego (approved):

PR #1481, opened 2026-08-04
[RW-1484] Add duplicate detection

orakili:
Refs: RW-1484

On report creation, this looks for duplicate document (similar body text) from the same source(s) posted within a week and mark the document as duplicate or to review.

When available, this extends the search for duplicates using embeddings to find same documents posted by other sources or in the past.

This requires version 1.2.0 of the OCHA AI Helper for the embedding endpoints. Currently it uses MariaDB VECTOR functionalities to store the embeddings because it's fast, doesn't require another store and allows to optimize filtering without duplicating data. This could be replaced with a different backend by implementing a new backend service.


cafuego (approved):
😬

PR #1488, opened 2026-08-20
[RW-1535] Add optional scheme, authentication, TLS verification settings for the elasticsearch/opensearch connection

orakili:
Refs: RW-1535


cafuego (approved):

cafuego (approved):
Do it!

